PR #27, opened 2023-09-05
subscribe action + simulate action | WIP

guitavano:
I didn't find an endpoint at storefront API, so I had to create a new GraphqlClient connecting to admin API, follow the structure that gimenes did before.

CustomerCreate at storefront: https://shopify.dev/docs/api/storefront/2023-07/mutations/customerCreate

It requires emails and password, I think it's use to create logins.

CustomerCreate at admin api: https://shopify.dev/docs/api/admin-graphql/2023-07/mutations/customerCreate

Buttt, the new client requires a new token. In this PR I just created a normal string field to receive that at admin.

What do you think about it? Do we have to use Secrets?

### DIFF
--- a/shopify/actions/newsletter/subscribe.ts
+++ b/shopify/actions/newsletter/subscribe.ts
@@ -1,5 +1,5 @@
 import { AppContext } from "../../mod.ts";
-import { Data, Variables, query } from "../../utils/queries/subscribe.ts";
+import createCustomerAction from "../customer/createCustomer.ts";
 
 type SubscribeCustomerProps = {
   input: {
@@ -13,29 +13,18 @@ type SubscribeCustomerProps = {
   };
 };
 
+
 const action = async (
   { input }: SubscribeCustomerProps,
-  _req: Request,
+  req: Request,
   ctx: AppContext,
 ): Promise<{id: string}> => {
-  const { admin } = ctx;
 
-  const { payload : {customer} }  = await admin.query<Data, Variables>({
-    variables: { 
-      input:{
-        email: input.email,
-        emailMarketingConsent: {
-          marketingState: input.emailMarketingConsent.marketingState,
-          consentUpdatedAt: input.emailMarketingConsent.consentUpdatedAt,
-          marketingOptInLevel: input.emailMarketingConsent.marketingOptInLevel
-        },
-        tags: input.tags
-      }
-    },
-    query,
-  });
+  const { id } = await createCustomerAction({ input }, req, ctx)
 
-  return customer;
+  return {
+    id
+  };
 };
 
 export default action;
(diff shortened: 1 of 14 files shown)


shopify/actions/newsletter/subscribe.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const action = async (
+  { input }: CreateCustomerProps,
+  _req: Request,
+  ctx: AppContext,
+): Promise<{id: string}> => {
+  const { admin } = ctx;
+  
+  console.log(input)
```
lucis, line 23:
```suggestion
```

shopify/actions/newsletter/subscribe.ts
```diff
@@ -0,0 +1,43 @@
+import { AppContext } from "../../mod.ts";
+import { Data, Variables, query } from "../../utils/queries/subscribe.ts";
+
+type CreateCustomerProps = {
```
lucis, line 4: Is the name right?

shopify/utils/queries/subscribe.ts
```diff
@@ -0,0 +1,35 @@
+import { gql } from "../../../utils/graphql.ts";
+
+export const query = gql`
+mutation customerCreate($input: CustomerInput!) {
+  payload: customerCreate(input: $input) {
```
lucis, line 5: The API for subscribing a customer on Shopify is called `customerCreate`?

shopify/mod.ts
```diff
@@ -18,6 +18,12 @@ export interface Props {
    */
   storefrontAccessToken: string;
 
+  /**
+   * @ttile Access Token
+   * @description Shopify admin access token.
+   */
+  adminAccessToken: string;
```
lucis, line 25: Shouldn't you use Secret here?

shopify/actions/newsletter/subscribe.ts
guitavano: I think SubscribeCustomerProps is better, CreateCustomer can be used in the futurem. thanks

shopify/mod.ts
guitavano: Yes, but Secrets are promises and today I can't use there, I will talk to candeia about it.

shopify/utils/queries/subscribe.ts
guitavano: Kind of that. 

They use the same table of Customers to do E-mail campaigns, the thing is that some customers accepted e-mail:
image
And each customer is tagged with specific campaigns:
image

But, you made me realize that I am not dealing with people that already are in the table. Now I will.

shopify/mod.ts
guitavano: I am using now

guitavano:
Not ready.

PR #43, opened 2023-09-12
VNDA-CEP-PDP

gabrielMatosBoubee:
Hello, here you can test the cep in simulation action. You need to pass SkuId, zip (PostalCode) and the quantity. To test it you can import in shipping simulate and take a look in array.

I don't decide what key i will put, so I need some ideas.

### Response EX: 
`{
    "": [
        {
            "name": "Retirada em Loja Moema",
            "value": "normal",
            "price": 0,
            "description": "Prazo de até 2 dias corridos para a entrega do pedido",
            "delivery_days": 2,
            "fulfillment_company": "intelipost",
            "value_needed_to_discount": null,
            "shipping_method_id": null,
            "notice": null,
            "countries": null
        },
        {
            "name": "Total Express",
            "value": "expressa",
            "price": 29.53,
            "description": "Prazo de até 4 dias corridos para a entrega do pedido",
            "delivery_days": 4,
            "fulfillment_company": "intelipost",
            "value_needed_to_discount": null,
            "shipping_method_id": null,
            "notice": null,
            "countries": null
        }
    ]
}`


vnda/actions/cart/simulation.ts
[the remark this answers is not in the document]
matheusgr: Also, better call it as shippingSimulation.

matheusgr (changes requested):

vnda/actions/cart/simulation.ts
```diff
@@ -0,0 +1,27 @@
...
+const action = async (
+  props: Props,
+  _req: Request,
+  ctx: AppContext,
+): Promise<ShippingMethod[]> => {
+  const { api } = ctx;
+  const {skuId, quantity, zip} = props
+  
+  if (!skuId || !quantity || !zip) {
+    throw new HttpError(400, "Missing cart cookie");
```
mcandeia, line 20: use `import { badRequest } from "deco/mod.ts";`

then just use

badRequest({message: "missing cart cookie"}) // or something like that

vnda/hooks/useCart.ts
```diff
@@ -47,6 +47,7 @@ const state = {
   update: action("vnda/actions/cart/updateCart.ts"),
   addItem: action("vnda/actions/cart/addItem.ts"),
   updateItem: action("vnda/actions/cart/updateItem.ts"),
+  simulate: invoke.vnda.actions.cart.simulation,
```
mcandeia, line 50: shouldn't this be wrapped as well?
```suggestion
  simulate: action("vnda/actions/cart/simulation"),
```

mcandeia (changes requested):

PR #46, opened 2023-09-13
Add support for searching by productId on productList loader

ingridflack:
## Description

This PR aims to add support for [searching by productId](https://developers.vtex.com/docs/api-reference/search-api#get-/api/catalog_system/pub/products/search) on the productList loader by adding 2 new props while keeping backward compatibility with the `ids` prop.

## Changes

- Add `skus` and `productIds` props that will have the data for each type of search
- The `ids` prop fallbacks to the same behavior as `skus` to be backward compatible
- Add a helper function to avoid code duplication

## Test instructions

- Test using `ids` or `skus`, should return a list of skus;
- Test using `productIds` should return a list of products;
- Other behaviors should stay the same.

### DIFF
--- a/vtex/loaders/legacy/productList.ts
+++ b/vtex/loaders/legacy/productList.ts
@@ -49,11 +49,16 @@ export interface FQProps extends CommonProps {
   count: number;
 }
 
-export interface ProductIDProps extends CommonProps {
+export interface SkuIDProps extends CommonProps {
   /**
    * @description SKU ids to retrieve
    */
-  ids: ProductID[];
+  ids?: ProductID[];
+  skus?: ProductID[];
+}
+
+export interface ProductIDProps extends CommonProps {
+  productIds?: ProductID[];
 }
 
 export interface CommonProps {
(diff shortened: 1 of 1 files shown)


vtex/loaders/legacy/productList.ts
```diff
@@ -88,10 +102,24 @@ const fromProps = (
     O?: LegacySort;
   };
 
+  if (isSKUIDProps(props)) {
+    const ids = props.ids || [];
+    const skus = props.skus || [];
+
+    const skuIds = new Set([...ids, ...skus]);
+    Array.from(skuIds).forEach((skuId) => params.fq.push(`skuId:${skuId}`));
```
lucis, line 110: Why are you passing the `ids` and `skus`? Don't they represent different things?

ingridflack: I added the additional `skus` prop due to backward compatibility concerns but removed it on [250b1c4](https://github.com/deco-cx/apps/pull/46/commits/250b1c4089b597066530e02f7302210a5eef9631). I kept the original `ids` prop and improved the comments and typing.

lucis (approved):

PR #56, opened 2023-09-16
Allow category tree filters on vtex legacy filters

jovenan:
Feature description:
This PR Allows the category tree filters to Vtex Legacy ProductListingPage. By sending to front the categories below the current department or category.

Use case:
https://www.austral.com.br/masculino?O=OrderByScoreDESC
https://australroupas.deco.site/masculino?O=OrderByScoreDESC

### DIFF
--- a/vtex/loaders/legacy/productListingPage.ts
+++ b/vtex/loaders/legacy/productListingPage.ts
@@ -14,7 +14,7 @@ import {
 } from "../../utils/segment.ts";
 import { withIsSimilarTo } from "../../utils/similars.ts";
 import { legacyFacetToFilter, toProduct } from "../../utils/transform.ts";
-import type { LegacyProduct, LegacySort } from "../../utils/types.ts";
+import type { LegacyProduct, LegacySort, LegacyFacet } from "../../utils/types.ts";
 
 const MAX_ALLOWED_PAGES = 500;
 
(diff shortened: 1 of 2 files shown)


jovenan:
This PR is the same of the STD PR: https://github.com/deco-sites/std/pull/179 however without correcting the URL that returns the same page as this feature has already been corrected

vtex/loaders/legacy/productListingPage.ts
```diff
@@ -177,8 +178,33 @@ const loader = async (
       props.similars ? withIsSimilarTo(req, ctx, product) : product
     ),
   );
+
+  // Get categories of the current department/category
+  const getCategoryFacets = (CategoriesTrees: LegacyFacet[]): LegacyFacet[] => {
+    const isDepartmentOrCategoryPage = !pageType;
```
lucis, line 184: Is this condition accurate?

jovenan: yes, because if the page is not a departament or category, an error will occur

jovenan: @lucis

tlgimenes (approved):
I have mixed feelings about this one. The loader should "load" the data and it's up to the react component to filter it and render it. This implementation makes it impossible for someone to render other categories because the loader is filtering the data. 

I'll accept since this loader may be deprecated in the future, but a nit would be to return all data an filter it on the frontend

PR #57, opened 2023-09-18
add simulate shipping

guitavano:


mcandeia (approved):

commerce/utils/stateByZip.ts
```diff
@@ -0,0 +1,49 @@
+const getStateFromZip = (cep: string) => {
```
mcandeia, line 1: Giving that we likely will have stores from outside brazil, I think we should adopt an approach that let's the user choose from where the state should came from. 

Given that, I think we should create an interface to properly accommodate the idea of having multiples "AddressLocator" and those "AddressLocator" receives a Zip as a parameter and return an "Address" as a response. This one is specifically used for Brazil but there are other for the US for instance.

So let's the user choose between then and even having the possibility to use more than one. 

One option is to modify your `AppContext` to have a new property `.addressLocator` which has the following interface

```ts
interface AddressLocator {
    byZipCode(zip: string): Promise<Address>
}
```

mcandeia (approved):

mcandeia (approved):

PR #60, opened 2023-09-18
Tavano/fix seo plp search

guitavano:


vtex/utils/legacy.ts
```diff
@@ -96,6 +96,17 @@ export const pageTypesToBreadcrumbList = (
 export const pageTypesToSeo = (pages: PageType[], req: Request): Seo | null => {
   const current = pages.at(-1);
 
+  const url = new URL(req.url);
+  const urlParams = new URLSearchParams(url.search);
```
lucis, line 100: I think `url.searchParams` is the same thing

lucis (approved):

PR #61, opened 2023-09-18
Allow passing `pageOffSet` as props to IS productListingPage

RafaelCassiano30011:
I made a modification to check if there is a current page being passed as a prop. Previously, we were only retrieving the "page" parameter from the URL, which worked only when the page was refreshed. The code was like this:

```
const page =
  Math.min(
    url.searchParams.get("page") ? Number(url.searchParams.get("page")) - currentPageoffset : 0,
    VTEX_MAX_PAGES - currentPageoffset
  );

```
Now, we have made the following change:

```
const page =
  props.pageOffset ??
  Math.min(
    url.searchParams.get("page") ? Number(url.searchParams.get("page")) - currentPageoffset : 0,
    VTEX_MAX_PAGES - currentPageoffset
  );
```
Now, the code checks if the "pageOffset" property is present in the props. If it's present, it will use that value; otherwise, it will continue with the previous logic to obtain the page from the URL.

### DIFF
--- a/vtex/loaders/intelligentSearch/productListingPage.ts
+++ b/vtex/loaders/intelligentSearch/productListingPage.ts
@@ -1,38 +1,16 @@
+/** @format */
+
 import type { ProductListingPage } from "../../../commerce/types.ts";
 import { parseRange } from "../../../commerce/utils/filters.ts";
 import sendEvent from "../../actions/analytics/sendEvent.ts";
 import { AppContext } from "../../mod.ts";
-import {
-  toPath,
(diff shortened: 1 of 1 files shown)


vtex/loaders/intelligentSearch/productListingPage.ts
```diff
@@ -1,38 +1,16 @@
+/** @format */
```
mcandeia, line 1: remove

mcandeia (approved):
fix then ship it!

PR #62, opened 2023-09-18
feat: Add missing pricerange filter for vtex legacy

IncognitaDev:


vtex/utils/transform.ts
```diff
@@ -543,13 +545,17 @@ export const legacyFacetToFilter = (
...
   return {
     "@type": "FilterToggle",
     quantity: facets.length,
     label: name,
     key: name,
     values: facets.map((facet) => {
+      if (name == "PriceRanges") {
+        facet.Map = "priceFrom";
+        facet.Value = facet.Slug!;
+      }
```
mcandeia, line 557: I would prefer if you return a new facet instead of mutating the original, how that's sound?

mcandeia (approved):
Fix then shipit!

vtex/utils/transform.ts
IncognitaDev: Its realy better, done.

PR #83, opened 2023-09-28
Add analytics app.

matheusgr:


tlgimenes (approved):

analytics/components/Jitsu.tsx
```diff
@@ -1,7 +1,31 @@
...
+  async?: boolean;
+}
+
+interface SectionProps {
+  async?: boolean;
+  domain: string;
+  path: string;
+}
+
+// TODO: WIP
```
igorbrasileiro, line 17: Remove

analytics/components/Jitsu.tsx
```diff
@@ -1,7 +1,31 @@
-import { context } from "deco/live.ts";
+import { Head } from "$fresh/runtime.ts";
```
igorbrasileiro, line 1: Is this a breaking change, isn't?

analytics/components/Plausible.tsx
```diff
@@ -0,0 +1,104 @@
...
+              data-exclude="/proxy"
+              data-api="https://plausible.io/api/event"
+              dangerouslySetInnerHTML={{
+                __html: plausibleScript,
+              }}
+            />
+          )}
+        {!async && (
+          <script
+            defer
```
igorbrasileiro, line 85:
```suggestion
            type="module"
```
defer doesn't work withou src https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer image

analytics/components/Plausible.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <Head>
+        <link rel="dns-prefetch" href="https://plausible.io/api/event" />
+        <link
+          rel="preconnect"
+          href="https://plausible.io/api/event"
+          crossOrigin="anonymous"
+        />
+        {async &&
+          (
+            <Script
```
igorbrasileiro, line 74: This Script that render Head, inside another Head, wasn't working, have you tested?

analytics/components/Plausible.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const plausibleScript = exclusionScript;
+
+const sendEvent = (
```
igorbrasileiro, line 28: Add a comment, that this function should be self contained, because it is `Stringified`

analytics/mod.ts
```diff
@@ -0,0 +1,13 @@
+import type { App, AppContext as AC } from "deco/mod.ts";
+import manifest, { Manifest } from "./manifest.gen.ts";
+
+export type AppContext = AC<ReturnType<typeof App>>;
+
+// deno-lint-ignore no-explicit-any
+export type State = any;
```
igorbrasileiro, line 7: It can be null?

analytics/sections/Analytics/Jitsu.tsx
```diff
@@ -0,0 +1,16 @@
+import Jitsu, { Props as JitsuProps } from "../../components/Jitsu.tsx";
+
+type Props = JitsuProps;
+
+function Section(props: Props) {
+  return <Jitsu domain={""} path={""} {...props} />;
```
igorbrasileiro, line 6:
```suggestion
  return <Jitsu domain="" path="" {...props} />;
```

analytics/sections/Analytics/Jitsu.tsx
```diff
@@ -0,0 +1,16 @@
+import Jitsu, { Props as JitsuProps } from "../../components/Jitsu.tsx";
+
+type Props = JitsuProps;
```
igorbrasileiro, line 3:
```suggestion
import Jitsu, { Props } from "../../components/Jitsu.tsx";
```
Does it make sense?

website/components/Analytics.tsx
```diff
@@ -0,0 +1,120 @@
...
+          src={src}
+          dangerouslyRunOnMainThread={dangerouslyRunOnMainThread}
+        />
+      )}
+
+      <Script
+        dangerouslySetInnerHTML={{
+          // add all globals variables here
+          __html:
+            `debugGlobals = () => { console.table([["datalayer", dataLayer]]); }`,
```
igorbrasileiro, line 105: Make sense add "plausible datalayer" here?

igorbrasileiro (approved):

analytics/components/Jitsu.tsx
matheusgr: This is really WIP :)

analytics/components/Jitsu.tsx
matheusgr: Not really since this component is not used outside the apps itself.

analytics/components/Plausible.tsx
matheusgr: I will remove async usage (not really used today)

matheusgr: Removed from plausible.

analytics/mod.ts
matheusgr: no, but Schema will not be mergeable if we specify it as a Record<string | number, never>. set it type as {} would also be strange.

analytics/sections/Analytics/Jitsu.tsx
matheusgr: yes!

website/components/Analytics.tsx
matheusgr: not really. I would avoid to be specific to something other than GA/GTM. we want only the analytics that almost every website will use.

PR #85, opened 2023-09-29
Bump deco to 1.37.6.

matheusgr:


matheusgr:
(can be bumped later)

PR #89, opened 2023-10-02
Fix: vtex search params IS/Legacy

IncognitaDev:
###  Search with path / Empty search in vtex IS

When a url path does not contain `s` query and its not a page/category/department the return was a seach page with all products. To fix, I add the default vtex behavior, any path that is not a page its coverted by a search query, so when its a not valid query the return is a not found (if implemented by storefront) and when is, the return is the search return.

### Search with path in vtex Legacy

As said above, a path thats is no a page its a valid search term, the behavior is the same.

### DIFF
--- a/vtex/loaders/intelligentSearch/productListingPage.ts
+++ b/vtex/loaders/intelligentSearch/productListingPage.ts
@@ -255,10 +255,11 @@ const loader = async (
   const url = new URL(baseUrl);
   const segment = getSegment(req);
   const currentPageoffset = props.pageOffset ?? 1;
-  const { selectedFacets: baseSelectedFacets, page, ...args } = searchArgsOf(
-    props,
-    url,
-  );
+  const {
+    selectedFacets: baseSelectedFacets,
+    page,
+    ...args
+  } = searchArgsOf(props, url);
   const pageTypesPromise = pageTypesFromPathname(url.pathname, ctx);
   const selectedFacets = baseSelectedFacets.length === 0
     ? filtersFromPathname(await pageTypesPromise)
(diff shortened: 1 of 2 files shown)


vtex/loaders/intelligentSearch/productListingPage.ts
```diff
@@ -250,18 +250,41 @@ const loader = async (
...
+  /**
+   *  verify if when url its not a category/department/collection
+   *  and is not a default query format but is a valid search based in default IS behavior on native stores
+   */
+  if (
+    !pageTypes.length &&
+    !selected.length &&
+    !args.query &&
+    pathList.length == 1
+  ) {
```
lucis, line 278: Create a variable with this condition and name it appropriately

vtex/loaders/intelligentSearch/productListingPage.ts
```diff
@@ -250,18 +250,41 @@ const loader = async (
...
+    args.query = pathList[0];
```
lucis, line 279: Can you avoid mutating the object and create a new const?

IncognitaDev: I also think that mutating its not good, I just cant think in a name for new args const

vtex/loaders/intelligentSearch/productListingPage.ts
IncognitaDev: done.

vtex/loaders/intelligentSearch/productListingPage.ts
IncognitaDev: done

lucis (approved):

PR #94, opened 2023-10-03
Fix vnda seo plp

gabrielMatosBoubee:
## Description

When I enter in category page, the HEADING, didn't appear. I change the order in the ternary and remove the seo api calls, because I didn't find nothing about that api in VNDA API. (https://developers.vnda.com.br/reference/refer%C3%AAncia-de-api)

Example: https://velocita.cdn.vnda.com.br/esp-tenis-finalidade-corrida

Now ours page follow this example

## File

vnda/loaders/productListingPage.ts


vnda/loaders/productListingPage.ts
```diff
@@ -63,31 +65,32 @@ const searchLoader = async (
...
+  const categoryTag = isSearchPage
+    ? undefined
+    : await api["GET /api/v2/tags/:name"]({ name: categoryTagName }, STALE)
+      .then((res) => res.json()).catch(() => undefined);
+
   const response = await api["GET /api/v2/products/search"]({
     term,
     sort,
     page,
     per_page: count,
```
matheusgr, line 79: To thos two calls at same time.

```suggestion
  const categoryTagPromise = isSearchPage
    ? Promise.resolve(undefined)
    : api["GET /api/v2/tags/:name"]({ name: categoryTagName }, STALE)
      .then((res) => res.json()).catch(() => Promise.resolve(undefined));

  const [categoryTag, response]= await Promise.all(categoryTagPromise, api["GET /api/v2/products/search"]({
    term,
    sort,
    page,
    per_page: count,
```

vnda/loaders/productListingPage.ts
```diff
@@ -63,31 +65,32 @@ const searchLoader = async (
...
-    "tags[]": props.tags,
+    "tags[]": props.tags && props.tags?.length > 0
+      ? props.tags
+      : (categoryTag?.name && props.filterByTags
+        ? [categoryTag?.name]
+        : undefined),
```
matheusgr, line 84: This seems a little confusing...

create some variables before that call:

```
const initialTags = props.tags && props.tags?.length > 0 ? props.tags : undefined;
const categoryTagsToFilter = (categoryTag?.name && props.filterByTags
        ? [categoryTag?.name]
        : undefined)
```

and use it to filter tags[]:

```suggestion
    "tags[]": initialTags ?? categoryTagsToFilter,
```

vnda/loaders/productListingPage.ts
```diff
@@ -63,31 +65,32 @@ const searchLoader = async (
...
-    api["GET /api/v2/seo_data"]({
-      resource_type: "Tag",
-      code: categoryTagName,
-      type: "category",
-    }, STALE).then((res) => res.json()),
-    isSearchPage
-      ? api["GET /api/v2/tags/:name"]({ name: categoryTagName }, STALE)
-        .then((res) => res.json()).catch(() => undefined)
-      : undefined,
   ]);
```
matheusgr, line 94: Don't need promise all here.

`const search = await response.json();`

matheusgr (changes requested):

vnda/loaders/productListingPage.ts
matheusgr: ignore it, cannot do categoryTag in paralel.

vnda/loaders/productListingPage.ts
```diff
@@ -63,34 +64,67 @@ const searchLoader = async (
...
+  const seoCategories = tags.slice(-[...categoryTagName].length).filter((
+    tag,
+  ): tag is { name: string; title: string; subtitle: string } =>
+    tag !== undefined
+  );
+
+  const categories = seoCategories.map(({ name }) => name);
+
+  const resolvedTagNames = tags
+    .filter((tag): tag is { name: string } => tag !== undefined)
```
matheusgr, line 94: prepare if tag has no name.

```suggestion
    .filter((tag): tag is { name?: string } => tag !== undefined && tag.name)
```

matheusgr (approved):

PR #98, opened 2023-10-03
feat: product list return kitlook array

jonathanwlopes:
Here we need an array with all kitlook products

The isKit property was added to the loader because this mapping needs to go through each item in the list, and in stores that do not use kitlook it will not be necessary to activate

When mapped and a kitlook product is found, the main product will have an array of kitItems

This array has the kit products, with the standard product information

Reason: In some cases the retailer needs to remove items from stock individually from the kit created on the platform, with this array we can handle item by item in the frontend, previously when used this loader only had the entire kit product without its individual items


vtex/loaders/intelligentSearch/productList.ts
```diff
@@ -127,17 +133,58 @@ const loader = async (
...
+            ? await vcsDeprecated
+              ["GET /api/catalog_system/pub/products/search/:term?"]({
+                ...params,
+                fq: product.items[0].kitItems.map((item) =>
+                  `skuId:${item.itemId}`
+                ),
+              }, { deco: { cache: "stale-while-revalidate" } }).then((
+                res: Response,
+              ) => res.json())
+            : [];
```
lucis, line 174: You're doing multiple requests here when you could batch it

Please, follow the `withIsSimilarTo` function down below that you'll find something already implemented for this.

lucis (changes requested):

vtex/loaders/intelligentSearch/productList.ts
jonathanwlopes: I refactored the code.

Now we have a single source of truth to return products with kitlook, facilitating implementation in other layers after this has been agreed.

The API: /api/io/_v/api/intelligent-search/product_search/*facets  it does not return kit information, which is why when you activate the kit option we look for the products it contains in the legacy API

The legacy API itself returns if the product is a kit and if it is the ids of the items that make up the kitlook

Therefore, I make a request to have these products and format in schema.org standards

The return we have is:

image

If the product does not have a kit, the return will remain unchanged.

With this data in the product, we can manipulate kit products on shelves on any page, as well as send each kit item to checkout

Thanks.

tlgimenes:
related to

PR #108, opened 2023-10-06
fix: googleFonts loader

igorbrasileiro:
### Issue
The loader isn't fetching on google the write font, based on user-agent headers and isn't replacing all urls from stylesheet.

Before: returning .ttf font using chrome, and only 1 because user-agent isn't considered.
image

After:
image


tlgimenes (approved):
AHTM

PR #116, opened 2023-10-10
Fix vtex plp loader

gabrielMatosBoubee:
## Problem

image

Before, in the loader (VTEX - PLP - LEGACY), the term in the link within the request was "/lancamento/Tinta-de-cabelo." Now, in the solution, the term is transformed into "/138/Tinta-de-cabelo." This change only applies to categories that are part of a product cluster.

## Solution

image


tlgimenes:
merged on #117

PR #120, opened 2023-10-11
feat: implementing option to inject kit items in product list

Brunight:
This PR aims to enable injecting VTEX's kit items into Intelligent Search's `productList` loader with full product data, not only the ids, using batch requests to the Catalog/Legacy API.
The kit items will be named as `isAccessoryOrSparePartFor`, just like Deco did in the `productDetailsPage` loader.

First the loader does the regular `productList` request to Intelligent Search's API, filters all the `productGroupIDs`, does a single request to Catalog/Legacy API that return all the kit items skus for all products, and finally calls the `productList` request again with those skus. Then it maps each initial product injecting its kit items as `isAccessoryOrSparePartFor`.

This PR was initially made by @jonathanwlopes (#98).

### DIFF
--- a/vtex/loaders/intelligentSearch/productList.ts
+++ b/vtex/loaders/intelligentSearch/productList.ts
@@ -6,6 +6,7 @@ import {
   withDefaultFacets,
   withDefaultParams,
 } from "../../utils/intelligentSearch.ts";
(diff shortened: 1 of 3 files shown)


tlgimenes:
Thanks for your contribution @Brunight !

Closing since it's now merged into #119

PR #127, opened 2023-10-12
Adapting the Plausible script to track segmented checkout routes.

nathanluiz33:
Used parameter 'script.hash' when generating the script on plausible. Still had to adapt the script as described on apps/utils/plausible_scripts.ts.


lucis (approved):
Sent direct feedback

nathanluiz33:
Updated Plausible script extension to also identify hash-based navigation. We used the script:
https://plausible.io/js/script.hash.exclusions.js

Here it has the reference to all possible extensions: https://plausible.io/docs/script-extensions. Also, there were some modifications on the script described on 'utils/plausible_scripts.ts'.

PR #128, opened 2023-10-13
Set segment only when cookie diverges

mcandeia:


vtex/middleware.ts
```diff
@@ -1,23 +1,39 @@
...
+/**
+ * by default segment starts with null values
+ */
+const DEFAULT_SEGMENT: Partial<Segment> = {
+  utmi_campaign: null,
+  utm_campaign: null,
+  utm_source: null,
+};
+
+const SEGMENT_ONCE_KEY = Symbol("Set segment on context");
```
lucis, line 20: Why do we need `Symbol` here? I remember briefly

Either way, WDYT about adding a comment about it?

lucis (approved):
Do you need help testing it? I think Dudu from ZeeDog can help you getting a valid `utm_segment`

vtex/middleware.ts
mcandeia: We are using WeakMap for bag @lucis Symbols are the only possible way for indexing WeakMaps. 

https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

I can certainly add a comment about it linking to the `WeakMap` docs, how about that?

Anyways I think this should not be necessary as the type of bag is weakmaps which should be addressed with documentation

mcandeia: About testing: I think i did not change the behavior of segment itself, just not setting it twice when it's already set

PR #131, opened 2023-10-17
feat: Events API

tlgimenes:
This PR introduces our new event API. This new API allows for apps to hook into the event system sent by the store. We use this to allow algolia app to listen for events the store is sending and do its thing. 

## How it works?
I created a new `website/components/Events.tsx`. This component is added to `webiste/pages/Page.tsx` and extends an existing global variable to the browser's window. This object contains 2 methods:
1. Dispatch
2. Subscribe

The store uses the `dispatch` function. This function warns all subscribed listeners synchronously (to avoid loosing events). This global variable inside the window is added via an inline script tag. This script tag uses the base64 inline technique to use the `defer` attrribute. This allows parsing the page before running anything, delaying FCP and decreasing TBT. This script is rendered on top of the page, meaning it will add the variable before any other analytics script. 

> It's important that anyone wanting to listen for events to add their script the same way we do, a.k.a, with base64 tags and defer

A default listener is them added for Plausilble under the Analytics app


website/components/Events.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+type EventHandler = (event?: AnalyticsEvent) => void | Promise<void>;
+
+type EventsAPI = {
+  dispatch: (event: unknown) => void;
+  subscribe: (
+    handler: EventHandler,
+    options?: AddEventListenerOptions | boolean,
+  ) => () => void;
+};
```
igorbrasileiro, line 13: Wdyt about using interface?

website/components/Events.tsx
```diff
@@ -0,0 +1,64 @@
...
+  window.DECO = {
+    ...window.DECO,
+    events: { dispatch, subscribe },
+  };
+};
+
+function Events() {
+  return (
+    <Head>
+      <script defer id="deco-events" src={scriptAsDataURI(snippet)} />
```
igorbrasileiro, line 59: Why is needed to use URI? Is it to leverage from `script + defer` ?

tlgimenes: Yes, apparently, defer does not work with inline scripts

tlgimenes: Yes, but it doesn't guarantee the script execution order. Defer makes sure the scripts execute in the right order

igorbrasileiro: https://gist.github.com/jakub-g/385ee6b41085303a53ad92c7c8afd7a6
According to this gist, the type inline script with type module implies `defer`.

tlgimenes: I mean, 
```
<script src="script-1"  defer />
<script src="script-2"  defer />
```

I know that script-1 will run before script-2. With type="module" I do not have that certainty

PR #137, opened 2023-10-18
url path search match

IncognitaDev:
Vtex understands unique path that not are a department as a search. This PRs prevents searchs wihtout query params that match as category or subcategory to be passed as `category-1`(department) and use term as search.

This PR also fix search path search when the vtex page type return is fulltex.

### DIFF
--- a/vtex/loaders/intelligentSearch/productListingPage.ts
+++ b/vtex/loaders/intelligentSearch/productListingPage.ts
@@ -202,8 +202,13 @@ const queryFromPathname = (
 ) => {
   const pathList = path.split("/").slice(1);
 
-  const isPage = Boolean(pageTypes.length);
-  const isValidPathSearch = pathList.length == 1;
+  const isUniquePath = pageTypes.length == 1;
+
+  const isPage = pageTypes.length > 1 ||
+    (isUniquePath &&
+      !NotAllowedPageTypeAsUniquePath.includes(pageTypes[0].pageType));
+
+  const isValidPathSearch = isUniquePath;
 
   if (!isPage && !isInSeachFormat && isValidPathSearch) {
     // decode uri parse uri enconde symbols like '%20' to ' '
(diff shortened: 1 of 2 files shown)


vtex/loaders/intelligentSearch/productListingPage.ts
```diff
@@ -202,7 +202,11 @@ const queryFromPathname = (
 ) => {
   const pathList = path.split("/").slice(1);
 
-  const isPage = Boolean(pageTypes.length);
+  const isPage = pageTypes.length > 1 ||
+    (pageTypes.length == 1 &&
+      pageTypes[0].pageType !== "FullText" &&
+      pageTypes[0].pageType !== "Category" &&
+      pageTypes[0].pageType !== "SubCategory");
```
mcandeia, line 209: it is not clear to me why it should have length exactly 1. Can you extract in a variable and put a comment on top of it?

Also: I think it could be more readable if you extract the possibilities to a const and then use `includes`

like

`["FullText", "Category", "SubCategory"].includes(pageType)`
```suggestion
  const [firstPageThatYouShouldPutAgoodNameWhyItShouldBeTheFirstParameter, ...pageTypes] = pageTypes;
  const isPage = pageTypes.length > 0 || ["FullText", "Category", "SubCategory"].includes(firstPageThatYouShouldPutAgoodNameWhyItShouldBeTheFirstParameter);
```

vtex/loaders/intelligentSearch/productListingPage.ts
```diff
@@ -211,8 +215,15 @@ const queryFromPathname = (
   }
 };
 
-const filtersFromPathname = (pages: PageType[]) =>
-  pages
+const filtersFromPathname = (pages: PageType[]) => {
+  if (
+    pages.length === 1 &&
+    (pages[0].pageType === "Category" || pages[0].pageType === "SubCategory")
```
mcandeia, line 221: can you extract the `Category` and `SubCategory` to a const and reuse it?

```ts
const pagesCategoryType = ["Category", "SubCategory"];
const wellKnownPageTypes = [...pagesCategoryType, "FullText"]; 
/// code
```

mcandeia (changes requested):

guitavano:
Closed by inactivity. If it is still happening, reopen.

PR #146, opened 2023-10-20
feat: kititems in loaders

jonathanwlopes:
My greetings,

The objective of this pullrequest is to bring kititem information for VTEX products,
By default this information should be part of the product
However, they are not returned in intelligent search loaders

To make it possible to access this important data, I added a request to get this data
After retrieving the product's kititem information, it is added to the 'isAccessoryOrSparePartFor' property
This occurs in the product list loader and also on the product listing page

With this information, through the front end, we can use this data to send separate products to the cart if it is a kit product, the benefit of this is to remove each product individually from stock.

See the return in productList
image

See the return in productListPage
image

If the product does not have a kit, an empty array is returned, see
image

By default, VTEX does not return kit information in the intelligent search API, until this issue is resolved, which would be the best of all worlds, this would be a solution for us to have this access

This information will be extremely useful in several use cases, I await evaluation


tlgimenes:
Issues I see in this one:

1. It uses Graphql API which uses post and thus, is NOT cacheable. This will increase a lot fetch times for Intelligent Search API. Currently our caching layer decreases this API latency from 1s to 40ms. This PR will make this latency go back to 1s.
2. For all stores that do not use kitItems, this will force fetching kit items and adding the latency.

What do you think about: 

1. Using a non graphql api
2. Adding a boolean attribute in props so we can activate kit items when needed?

guitavano:
Closed by inactivity. If it is still happening, reopen.

PR #147, opened 2023-10-20
Action add multiple items to cart and params from props on PLP

IncognitaDev:


mcandeia (changes requested):

wake/loaders/productDetailsPage.ts
```diff
@@ -46,6 +46,8 @@ async function loader(
     return null;
   }
 
+  console.log(wakeProduct);
```
mcandeia, line 49: remove

mcandeia (approved):
fix, then shipit

PR #170, opened 2023-10-26
feat: PLPs page numbers in canonical tags

AlanVncs:
From this change, it will be possible to decide whether the page number (PLPs) will be taken into account on indexing.

By default, the following pages would be indexed as the same page, that is, the same canonical tag would be generated for both:
`http://lojastorra.com.br/my-category?page=2`
`http://lojastorra.com.br/my-category?page=3`

However, when the user checks the `pgNumberIndexing` option, the above pages will be indexed separately.

### DIFF
--- a/commerce/sections/Seo/SeoPLP.tsx
+++ b/commerce/sections/Seo/SeoPLP.tsx
@@ -1,5 +1,6 @@
 import Seo, { Props as SeoProps } from "../../../website/components/Seo.tsx";
 import { ProductListingPage } from "../../types.ts";
+import { canonicalFromBreadcrumblist } from "../../utils/canonical.ts";
 
 export type Props = {
   jsonLD: ProductListingPage | null;
@@ -12,7 +13,7 @@ function Section({ jsonLD, ...props }: Props) {
     ? jsonLD.seo.canonical
     : jsonLD?.breadcrumb
     ? canonicalFromBreadcrumblist(jsonLD?.breadcrumb)
-    : null;
+    : props.canonical;
 
   const noIndexing = !jsonLD || !jsonLD.products.length;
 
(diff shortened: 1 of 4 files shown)


lucis (commented):
I think the API could accept the query param's names instead

website/components/Seo.tsx
```diff
@@ -29,6 +29,11 @@ export interface Props {
   themeColor?: string;
   /** @title Canonical URL */
   canonical?: string;
+  /**
```
tlgimenes, line 32: I think we don't need this. It's ok always to consider all pages on indexing, i.e. this should always be true

commerce/sections/Seo/SeoPLP.tsx
```diff
@@ -1,17 +1,25 @@
...
 import { ProductListingPage } from "../../types.ts";
-import { canonicalFromBreadcrumblist } from "../../utils/canonical.ts";
 
 export type Props = {
   jsonLD: ProductListingPage | null;
 } & Partial<Omit<SeoProps, "jsonLDs">>;
 
-function Section({ jsonLD, ...props }: Props) {
+export const loader = ({ ...props }: Props, req: Request) => {
+  return { ...props, url: req.url };
```
tlgimenes, line 9: I think we already have all necessary info inside jsonLD

tlgimenes (commented):
Nice addition! However, I think you changed on the wrong place. What about chaging the canonical in each loader instead? this way we can account for the pecualiarities in each platform. This way, we should change all PLP loaders to add this functionality. If you want, you can add this only on the loader you need

AlanVncs:
@lucis > I think the API could accept the query param's names instead
I think it's a great idea, but I also think that the page number should be a special property because it is a special parameter and has particularities in defining the canonical URL.

PR #177, opened 2023-10-27
feat: add script props to mantain retrocompatibility with std

IncognitaDev:
Add script props to mantain retrocompatibility with std version of component


guitavano:
@igorbrasileiro can you help us here?

igorbrasileiro (approved):

PR #182, opened 2023-10-30
fix canonicals

tlgimenes:
This PR uses #94 and:

1. fixes canonicals on plp, pdp
2. fixes filters on vnda


tlgimenes:
cheers to @gabrielMatosBoubee !

PR #199, opened 2023-11-07
feat: use secrets

tlgimenes:
migrate all apps to use secrets instead of plain string


lucis (approved):
Puzzled because `SecretString`, but ok

PR #200, opened 2023-11-08
feat: Caching loaders

tlgimenes:
Migrates VTEX to use the new traced loaders:

More info at: https://github.com/deco-cx/deco/pull/502

### DIFF
--- a/vtex/loaders/legacy/productList.ts
+++ b/vtex/loaders/legacy/productList.ts
@@ -194,4 +194,9 @@ const loader = async (
   );
 };
 
+export const cache = "stale-while-revalidate";
+
+export const cacheKey = (_req: Request, ctx: AppContext) =>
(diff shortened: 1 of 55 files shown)


vtex/loaders/legacy/productList.ts
```diff
@@ -194,4 +196,9 @@ const loader = async (
   );
 };
 
+export const cache = "stale-while-revalidate";
+
+export const cacheKey = (_req: Request, ctx: AppContext) =>
```
mcandeia, line 201: I'm noticing a few usages of the same pattern:

it might worth to have a 

export { cacheKey, cache } from "apps/vtex/caches/cacheBySegment.ts"; 

?

mcandeia (approved):

igorbrasileiro (approved):

PR #206, opened 2023-11-18
Enhancement: Inclusion of "Suggested Retail Price" in Product Details

breno-oliveira:
Submitting this pull request to ensure the preservation of the suggested retail price on products that undergo updates in their price table, as well as those featuring any enabled promotions on VTEX.

In the original product request on VTEX, there is a value for `PriceWithoutDiscount` that is currently typed but missing from the return provided by the product details loader. By adding this typed value as a Unit Price Specification for the suggested retail price, we can introduce a feature that displays the original price before any discounts from the price table, while still presenting the final price including all available promotions.

image

Example image taken from https://secure.zeedog.com.br/api/catalog_system/pub/products/search/?fq=productId:5052709

### DIFF
--- a/vtex/utils/transform.ts
+++ b/vtex/utils/transform.ts
@@ -502,6 +502,11 @@ const toOffer = ({ commertialOffer: offer, sellerId }: SellerVTEX): Offer => ({
       priceType: "https://schema.org/SalePrice",
       price: offer.Price,
     },
+    {
+      "@type": "UnitPriceSpecification",
+      priceType: "https://schema.org/SRP",
+      price: offer.PriceWithoutDiscount,
+    },
     ...offer.Installments.map(
       (installment): UnitPriceSpecification => ({
         "@type": "UnitPriceSpecification",


eduardobbrito:
Adding a bit more of context:

Zee.Dog features a subscription service that effectively works as a VTEX recurrence, incorporating a fixed discount through a promotional offer. Typically, this subscription discount stands as the sole discount applicable throughout the recurrence cycle. However, certain products may experience a price reduction specifically for Black Friday, achieved by updating the overall price table for the entire site.

A product in this context will have a `ListPrice` (derived from the undiscounted price table), a `Price Without Discount` that we're proposing in this pull request as `SuggestedPrice` (obtained from the discounted table, reflecting a, for instance, 5% reduction), and a `SalePrice` (which accumulates discounts from the price table along with any applicable promotions). Ultimately, there's a need to display all three of these prices.

<figure>
Screenshot 2023-11-17 at 22 20 21
<figcaption>desired outcome that needs `SuggestedPrice`</figcaption>
</figure>

mcandeia (approved):
LGTM

lucis (approved):

PR #207, opened 2023-11-20
feat: add vtex useSegment

devartes:
- adding useSegment related to: https://github.com/deco-sites/macoteste/pull/6


lucis (approved):
LGTM

I was under the impression that you'd only add the loader (which, for my surprise, doesn't contact VTEX, just parses the string), but you also added it to the object that performs fetches on page load (correct me if I'm wrong).

It seems inoffensive and a nice add.

tlgimenes:
I think we dont need this one.  It looks like macoteste does not need this info. Maybe this info is already available somewhere else for them

PR #211, opened 2023-11-22
feat: format unused-path at Page and Redirects

igorbrasileiro:


lucis (approved):

lucis:
Question: Shouldn't we make it a bit clearer that this means "path already present on the site", and not a random path?

Not sure if we're locking anything here

igorbrasileiro:
> Question: Shouldn't we make it a bit clearer that this means "path already present on the site", and not a random path?

I didn't get it what you mean, can you rephrase? As far as I understood what you said, the admin that handles this `format path-uri`

PR #214, opened 2023-11-22
Prefer canonical prop over response

matheusgr:
Canonical prop when set on SEO components will now be used instead of seo from loader.

It is not possible to currently override canonical from backend which can not be a desired effect on page.

### DIFF
--- a/commerce/sections/Seo/SeoPDP.tsx
+++ b/commerce/sections/Seo/SeoPDP.tsx
@@ -23,7 +23,7 @@ function Section({ jsonLD, ...props }: Props) {
       title={title || props.title}
       description={description || props.description}
       image={image || props.image}
-      canonical={canonical || props.canonical}
+      canonical={(props.canonical || canonical) ?? undefined}
       jsonLDs={[jsonLD]}
       noIndexing={noIndexing}
     />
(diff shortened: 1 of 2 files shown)


commerce/sections/Seo/SeoPDP.tsx
```diff
@@ -23,7 +23,7 @@ function Section({ jsonLD, ...props }: Props) {
       title={title || props.title}
       description={description || props.description}
       image={image || props.image}
-      canonical={canonical || props.canonical}
+      canonical={(props.canonical || canonical) ?? undefined}
```
tlgimenes, line 26:
```suggestion
      canonical={props.canonical || canonical}
```

commerce/sections/Seo/SeoPLP.tsx
```diff
@@ -9,11 +9,13 @@ export type Props = {
 function Section({ jsonLD, ...props }: Props) {
   const title = jsonLD?.seo?.title;
   const description = jsonLD?.seo?.description;
-  const canonical = jsonLD?.seo?.canonical
+  const canonical = props.canonical?.trim()
```
tlgimenes, line 12:
```suggestion
  const canonical = props.canonical
```

commerce/sections/Seo/SeoPLP.tsx
```diff
@@ -9,11 +9,13 @@ export type Props = {
...
+    ? props.canonical.trim()
```
tlgimenes, line 13:
```suggestion
    ? props.canonical
```

PR #218, opened 2023-11-23
Add agent id in cart VNDA

gabrielMatosBoubee:
## The feature

When you enter in any page with the cart, you want to create it if you don't have yet. In the creation, you need to pass to VNDA the "agent" (This info is for the affiliation win his commission). You can got this info in this example url: "URL/?agent=104"

## The code

I have created a middleware that, upon entering any page with this query string, will verify if a cookie already exists and if it matches. If the cookie is the same, it does nothing. However, if it is different, it creates a new one with a duration of 60 days. The cookie only has the agentId.

image

### DIFF
--- a/vnda/utils/segment.ts
+++ b/vnda/utils/segment.ts
@@ -1,0 +1,52 @@
+import { getCookies, setCookie } from "std/http/mod.ts";
+import { AppContext } from "../mod.ts";
+
+interface Segment {
+  agent: string;
+}
+
+export const SEGMENT_COOKIE_NAME = "vnda_segment";
+const SEGMENT = Symbol("segment");
+const SIXTYDAYS = new Date(Date.now() + 60 * 24 * 60 * 60 * 1000);
+
+export const getSegmentFromBag = (ctx: AppContext): string =>
+  ctx.bag?.get(SEGMENT);
+export const setSegmentInBag = (ctx: AppContext, segment: string) =>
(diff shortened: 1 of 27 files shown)


vnda/utils/segment.ts
```diff
@@ -0,0 +1,51 @@
...
+  segment: string,
+  headers: Headers = new Headers(),
+): Headers => {
+  setCookie(headers, {
+    value: segment,
+    name: SEGMENT_COOKIE_NAME,
+    path: "/",
+    secure: true,
+    httpOnly: true,
+    expires: SIXTYDAYS,
```
matheusgr, line 47: Could you please verify the expected behavior with a representative from Vnda?

My hypothesis is that it might be better a session cookie or a briefly-lasting cookie (perhaps 1 day).

matheusgr (approved):
please avoid updating/uploade not related manifests/files.

matheusgr (changes requested):
(see requests on the last review)

vnda/utils/segment.ts
gabrielMatosBoubee: Vnda has communicated that it is essential to keep this cookie active for a period of sixty days, a necessity dictated by the utilization of a tool called 'direct' within the Vnda platform.

The client, while currently don't have any purchase plans in this first days, could decide to buy something within this timeframe. In this case I need to create a cart with this agent.

matheusgr (approved):

PR #221, opened 2023-11-23
Fix: Missing `sc` segment param in VTEX legacy for sales channel

Brunight:
This PR aims to add a missing `sc` segment param in VTEX Legacy APIs. When updating the Sales Channel in Deco admin, the requests sent to VTEX Legacy APIs were still using the default Sales Channel `1`.


vtex/utils/legacy.ts
```diff
@@ -12,6 +12,7 @@ export const toSegmentParams = (
     utmi_campaign: segment.utmi_campaign ?? undefined,
     utm_campaign: segment.utm_campaign ?? undefined,
     utm_source: segment.utm_source ?? undefined,
+    sc: segment.channel ?? '1',
```
lucis, line 15: I think we have a `defaulSalesChannel` or something on the app. Could you check and maybe use this value?

Brunight: Isn't it from std? I couldn't find it inside the `vtex` app :/
Maybe I should leave it with `undefined` like the others?

mcandeia: LGTM,
@lucis any thoughts?

guitavano:
@Brunight is it necessary yet? Can you explain where you are using ``sc``?

Brunight:
Yes it is, we use it at [Eico](https://github.com/deco-sites/eico) as they still uses the legacy loader because they had issues with intelligent search (not Deco related)
Their vtex account has multiple sales channels, and without the fixes from this PR some core functionalities of the store just won't work, such as adding a product to cart, because the correct sales channel ID for the store is different from vtex's default when not providing it in query params

PR #223, opened 2023-11-29
Feat/store-review

pandero7:
## The feature 

This PR we use https://cl.avis-verifies.com/br/cache/8/6/a/${idWebsite}/AWS/WEBSITE_API/reviews.json You can consume this loader in deco like that 

```js
import type { Reviews } from "apps/commerce/types"

export interface Props { reviews: Reviews[] | null }
```

### DIFF
--- a/verified-reviews/loaders/storeReview.ts
+++ b/verified-reviews/loaders/storeReview.ts
@@ -1,0 +1,34 @@
+import { AppContext } from "../mod.ts";
+import { Review } from "../../commerce/types.ts";
+import { ExtensionOf } from "../../website/loaders/extension.ts";
+import {
+  createClient,
+  PaginationOptions,
+} from "../utils/client.ts";
+export type Props = PaginationOptions;
+
+/**
+ * @title Opiniões verificadas - Full Review for Product (Ratings and Reviews)
+ */
+export default function storeReview(
+  _config: Props,
+  _req: Request,
+  ctx: AppContext,
+): ExtensionOf<Review | null> {
+  const client = createClient({ ...ctx });
+  return async (storeReview: Review | null) => {
+    if (!storeReview) {
+      return null;
+    }
+
+    if (!client) {
+      return null;
+    }
+
+    const reviews = await client.storeReview();
+
+    return {
+        ...reviews,
+    };
+  };
+}
--- a/verified-reviews/loaders/storeReview.ts
+++ b/verified-reviews/loaders/storeReview.ts
@@ -1,3 +1,4 @@
+// deno-lint-ignore-file
 import { AppContext } from "../mod.ts";
 import { Review } from "../../commerce/types.ts";
 import { createClient } from "../utils/client.ts";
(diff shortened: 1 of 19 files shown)


verified-reviews/utils/client.ts
```diff
@@ -167,11 +167,50 @@ export const createClient = (params: ConfigVerifiedReviews | undefined) => {
...
 
+  const storeReview = async () : Promise<Review[] | null> => {
+    try {
+      const response = await fetchAPI<Reviews['reviews']>(`https://cl.avis-verifies.com/br/cache/8/6/a/${idWebsite}/AWS/WEBSITE_API/reviews.json`, {
+        method: "GET",
+      });
+
+      return(
+        response 
+        ?
```
mcandeia, line 178: I would avoid doing data transformations on clients. Clients should serve only as an interface for making http calls, the rationale behind this is because you can reuse clients across multiples actions and loaders without necessarily doing data transformations. 

If you can, move this code to the loader implementation instead

mcandeia (changes requested):

verified-reviews/utils/client.ts
pandero7: Sure. If I understood correctly, I fixed that now in this new commit, can you validate it for me, please?

verified-reviews/loaders/storeReview.ts
```diff
@@ -0,0 +1,54 @@
...
+    ],
+    datePublished: item.review_date,
+    reviewBody: item.review,
+    reviewRating: {
+      "@type": "AggregateRating",
+      ratingValue: Number(item.rate),
+    },
+  }))
+
+  const limitedReviews = formatReviews.slice(0, _config.limit);
```
mcandeia, line 51: You should use the pagination object instead https://github.com/deco-cx/apps/blob/6a14c13a2ffcd9696d382bb112139e5ccb6d0dfb/verified-reviews/utils/client.ts#L9, is it possible?

Using limit and slice its not so reliable since pagination has limits and you'll reach the limit faster than you think, so always use API pagination instead.

mcandeia (changes requested):

verified-reviews/loaders/storeReview.ts
guitavano: In general lines, @pandero7 can we do a fetch already sending the limit?

gabrielMatosBoubee: @guitavano @mcandeia In this end-point we don't have pagination. Any sugestion to put smth? Only remove slice?

verified-reviews/loaders/storeReview.ts
```diff
@@ -0,0 +1,54 @@
...
+import { AppContext } from "../mod.ts";
+import { Review } from "../../commerce/types.ts";
+import { createClient } from "../utils/client.ts";
+
+export type Props = {
+  /**
+   * @title Number of reviews 
+   */
+  limit: number;
+};
```
guitavano, line 11:
```suggestion
export type Props = {
  offset?: number;
  /**
   * @title Number of reviews 
   */
  limit?: number;
};
```

We know that api does not have pagination today, but we can do it at loader using offset.

verified-reviews/loaders/storeReview.ts
guitavano, line 51:
```suggestion
  const limitedReviews = formatReviews.slice(_config.offset || 0, _config.limit || formatReviews.length);
```

guitavano:
Continued here:  https://github.com/deco-cx/apps/pull/377/

PR #225, opened 2023-12-01
[VNDA] Preventing 'normalizeVariants' from returning an empty array

AlanVncs:
Issue:
When a `VNDAProductGroup` has no variants, `normalizeVariants` returns an empty array. It is a problem because `pickVariant` returns `undefined` (`head`) and it causes an error from accessing properties from `undefined`.

Solution:
The function `normalizeVariants` will return at least one VNDAProduct, assuming that when VNDAProductGroup has no variants, it means the product is unavailable.

### DIFF
--- a/vnda/utils/transform.ts
+++ b/vnda/utils/transform.ts
@@ -69,7 +69,7 @@ export const parseSlug = (slug: string) => {
 };
 
 const pickVariant = (product: VNDAProductGroup, variantId: string | null) => {
-  const variants = normalizeVariants(product.variants);
+  const variants = normalizeVariants(product);
   const [head] = variants;
 
   let [target, main, available]: Array<VNDAProduct | null> = [null, head, null];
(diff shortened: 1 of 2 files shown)


devartes:
I observed the addition of the configuration `"cSpell.words": ["norder"]` in the [settings.json](https://github.com/deco-cx/apps/pull/225/commits/8090508b435e07270dc93627b92fb793d41bd899#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357) file, specifically within the code snippet:

https://github.com/deco-cx/apps/pull/225/commits/8090508b435e07270dc93627b92fb793d41bd899#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357R5

I would like to understand the reasoning behind this addition, as it seems unnecessary, especially since the absence of cSpell did not generate errors related to "norder."

AlanVncs:
I removed the configuration.

In fact, considering the fact that the project does not have recommendations for "spell checker" extensions, I agree that the configuration is unnecessary/inconvenient.

vnda/utils/transform.ts
```diff
@@ -166,35 +168,93 @@ const toOffer = ({
...
-    isProductVariant(v) ? [v] : Object.values(v) as VNDAProduct[]
+// Converts a VNDAProductGroup with no variants into a VNDAProduct
+// *** If VNDAProductGroup has no variants, it is not available ***
+const buildNoVariantsProduct = (productGroup: VNDAProductGroup) => {
+  const {
+    id,
+    image_url = "",
+    installments = [],
+    name = "",
+    price = 0,
```
devartes, line 207: Regarding the price (price) returning as 0, I have doubts if this is the most appropriate choice. I would like an explanation on why returning 0 is considered ideal.

vnda/utils/transform.ts
```diff
@@ -166,35 +168,93 @@ const toOffer = ({
...
+    main: true,
+    min_quantity: 1,
+    name,
+    norder: 0,
+    price,
+    product_id: id!,
+    properties: {},
+    quantity: 0,
+    sale_price,
+    sku: "01",
```
devartes, line 233: Concerning the SKU, considering that using "01" may lead to conflicts, wouldn't it be better to leave it empty `sku: "",` ?

devartes:
pr replaced by this pr's solution: https://github.com/deco-cx/apps/pull/244

PR #237, opened 2023-12-11
VNDA - Use appropriate variant images

AlanVncs:
I implemented a flag (`useVariantImages`) that will change the way the images are attached to each variant.
Currently, all images are attached to all variants (witch is not right at all, I think). Anyway, with the flag set to `true`, each variant will be associated only with the appropriate images.

See the same product in both ways:

---

`useVariantImages = false`
`https://deco-sites-vip-capas.deno.dev/produto/produto-teste-10354?skuId=04`
image

---

`useVariantImages = true`
`http://localhost:8000/produto/produto-teste-10354?skuId=04`
image

### DIFF
--- a/vnda/loaders/productDetailsPage.ts
+++ b/vnda/loaders/productDetailsPage.ts
@@ -6,6 +6,7 @@ import { parseSlug, toProduct } from "../utils/transform.ts";
 
 export interface Props {
   slug: RequestURLParam;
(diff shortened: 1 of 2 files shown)


guitavano:
Steel need it? @AlanVncs

matheusgr:
No activity/answer in the last month.

PR #239, opened 2023-12-11
Create files app

mcandeia:


admin/actions/releases/fork.ts
```diff
@@ -0,0 +1,96 @@
...
+          metadata: {}, // TODO: add metadataß
+          ack,
+        });
+      } catch ({ name, operation }) {
+        console.error({ name, operation });
+      }
+    } else if (data.type === "fetch-state") {
+      send({
+        type: "state-fetched",
+        payload: await fetchState(),
```
mcandeia, line 69: should fetchstate return the revision as well?

tlgimenes (approved):

admin/actions/releases/fork.ts
tlgimenes: Yeah, let's see how the implementation outside localhost works. Maybe we'll need to change this again

PR #240, opened 2023-12-12
Feat:vnda/transform - additionalProperties from 'category_tags'

AlanVncs:
As you can [see here](https://gist.github.com/AlanVncs/67f37dd552fa6c872104f66c9a2abca1), VNDA API is returning a [Product](https://github.com/deco-cx/apps/blob/81a3dcfe0c6dd92a464aab1d353359ebccd0a674/vnda/utils/openapi/vnda.openapi.gen.ts#L3989), which has the property `category_tags`.
With the change in this PR, data from `category_tags` will be included in the final `additionalProperties`.

See the same product before and after changes:

`http://localhost:8000/produto/produto-teste-10354?skuId=01`

**Before**
image

**After**
image


gabrielMatosBoubee:
I had the same issue with VNDA. The category_tags were designed for PLP.

Their recommendation is that until they fix this API, you should use the following one: https://developers.vnda.com.br/reference/get-api-v2-products-search

I've already integrated it into Deco to pull by IDs, so an inline productList loader would solve your problem.

PR #255, opened 2023-12-20
Sitemap vnda

gabrielMatosBoubee:


vnda/loaders/proxy.ts
```diff
@@ -65,7 +65,9 @@ function loader(
     handler: {
       value: {
         __resolveType: "website/handlers/proxy.ts",
-        url: internalDomain,
+        url: pathTemplate === "/sitemap.xml"
+          ? `https://sitemap.vnda.com.br/preview/${publicUrl}`
+          : internalDomain,
```
mcandeia, line 70: I don't know if a direct proxy is the desirable here, very often the customer wants to include the deco's sitemap as well. Look how we do in the vtex site handler:  https://github.com/deco-cx/apps/blob/main/vtex/handlers/sitemap.ts#L36

mcandeia (changes requested):

PR #275, opened 2024-01-05
[Shopify] - Return SEO, collections and tags in PDP loader

vitoUwu:
This PR makes an edit to the product search Query to return the product collections, in addition, it also makes an edit to the "toProductPage" function to return the product collections, tags and SEO.


shopify/utils/transform.ts
```diff
@@ -224,8 +239,8 @@ const filtersURL = (filter: FilterShopify, value: FilterValue, _url: URL) => {
   params.delete("page");
   params.delete("startCursor");
   params.delete("endCursor");
-  if (params.has(filter.id, value.label)) {
-    params.delete(filter.id, value.label);
+  if (params.has(filter.id)) {
```
guitavano, line 242: I did not understand that change.

My case of use is:

Current URL = /clothes?filter.v.option.tamanho=G
Sizes at filter:
P = /clothes?filter.v.option.tamanho=G&filter.v.option.tamanho=P
M = /clothes?filter.v.option.tamanho=G&filter.v.option.tamanho=M
G = /clothes
GG = /clothes?filter.v.option.tamanho=G&filter.v.option.tamanho=GG

It is useful, because the second click is like to uncheck the option.

guitavano (approved):

shopify/utils/transform.ts
guitavano: Type error for him. Idk but back code.

PR #276, opened 2024-01-05
add ratingCount value

guitavano:


guitavano:
Addressed here: https://github.com/deco-cx/apps/pull/288

PR #278, opened 2024-01-08
[Website] - Fix deferred firing unnecessary requests

vitoUwu:
The "website" app has a component called "Deferred" that renders certain components based on the scroll or screen intersection. When we set it to render based on the scroll, the component executes several requests as we scroll until the first request finishes. To resolve this, I set the [`once`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once) property in the eventlistener, which causes the event callback be executed only once.

Without once: 
image

With once true:
image


tlgimenes (approved):
NICE

PR #317, opened 2024-01-17
fix canonical to lowercase

guitavano:
It should be default, who wants different, needs to customize.

### DIFF
--- a/vtex/utils/legacy.ts
+++ b/vtex/utils/legacy.ts
@@ -121,8 +121,6 @@ export const pageTypesToSeo = (
     return null;
   }
 
-  console.log(current.url)
-
   return {
     title: current.title!,
     description: current.metaTagDescription!,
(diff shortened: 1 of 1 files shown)


vtex/utils/legacy.ts
matheusgr, line 1: shouldn't also change line https://github.com/deco-cx/apps/pull/317/files#diff-546733263df1a40c27398f823e5d44e447b522d32f5b9e38635b02c8c94f14fbR116?

vtex/utils/legacy.ts
```diff
@@ -142,5 +142,5 @@ function toCanonical(url: URL, page?: number) {
     url.searchParams.set("page", `${page}`);
   }
 
-  return url.href;
+  return url.href.toLowerCase();
```
matheusgr, line 145: It seems reasonable to not lowercase searchparams.

matheusgr (approved):

vtex/utils/legacy.ts
guitavano: Actually, idk, let`s start w/ categories.

PR #332, opened 2024-01-22
[Shopify] - Add more information to product collections

vitoUwu:


guitavano:
I think change "name" can broke some frontend components that already use that. Can you send an example of this extra information added?

lucis:
@vitoUwu yes, exactly

If some frontend component, for example, does: `collections.filter(({ name }) => name === "COLLECTION"`, after this new version it will break.

What do you think about it?

vitoUwu:
@lucis I can't think of a good solution for both sides, if I changed "name" or "value" it would break the frontends that use the current way, and I needed the tilte (name) of the collection, its handle (value) and a way to identify that the object is a collection (propertyID). Do you have any suggestions on what can be done?

guitavano:
The most important thing is maintain current attributes.

what do you think about:

```js
{ 
"@type": "PropertyValue"
"@id": "gid/...."
name: "COLLECTION", 
value: "Lista de Bolsas",
valueReference: "listas"
}
```

vitoUwu:
Hmm, this should work

vitoUwu:
I kept the properties of the old object and added the new information in other properties

guitavano (approved):

PR #345, opened 2024-01-29
fix remove senseless props from app site

guitavano:
A lot of stores were having problem with canonical, when you fill canonical at app "site", it overlaps other ones, affecting PDPs and PLPs.

In general, I do not think that canonical make senses in app "site".

Enjoying the movement, i am removing titleTemplate and descriptionTemplate from app "site", because both not receive any data to replace.

### DIFF
--- a/website/mod.ts
+++ b/website/mod.ts
@@ -41,7 +41,7 @@ export interface Props {
   routes?: Routes[];
 
   /** @title Seo */
-  seo?: Omit<Seo, "jsonLDs">;
+  seo?: Omit<Seo, "jsonLDs" | "titleTemplate" | "descriptionTemplate" | "canonical">;
 
   /**
    * @title Global Sections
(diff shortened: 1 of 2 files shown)


website/mod.ts
```diff
@@ -41,7 +41,7 @@ export interface Props {
   routes?: Routes[];
 
   /** @title Seo */
-  seo?: Omit<Seo, "jsonLDs">;
+  seo?: Omit<Seo, "jsonLDs" | "titleTemplate" | "descriptionTemplate" | "canonical">;
```
lucis, line 44: Wouldn't `Pick<Seo, whatever` make more sense here?

lucis (approved):

website/mod.ts
guitavano: After your comment I tested it, but it did not work with Pick, @mcandeia is it mapped?

Anyway, I will merge to not loose the improvement.

PR #352, opened 2024-01-31
Tavano/fix image loader

guitavano:


guitavano:
We should not hide errors

PR #361, opened 2024-02-02
show more loader

guitavano:


vnda/utils/openapi/vnda.openapi.gen.ts
```diff
@@ -1388,7 +1388,9 @@ wildcard?: boolean
 /**
  * Filtra pelo nome da tag dentro de um tipo de tag. Exemplo, type_tags[cor]=verde
  */
-"type_tags[]"?: { key?: string; value?: string }[]
+"type_tags[]"?: {
```
lucis, line 1391: Hmm, I think this is being replaced by the auto-gen?

Should we do something to stop it being replaced

PR #365, opened 2024-02-02
[FIX] - VTEX MASTER DATA

gabrielMatosBoubee:
## The solution

Add the possibility to use private entities.

## The explain to use this solution

image

image

image


guitavano:
Do not merge before @mcandeia and @tlgimenes approve

mcandeia:
@guitavano isn't this a breaking change that we tried to do few weeks ago? The edge case is when the passed vtex headers are invalid and the masterdata entity is **public** the create should work as it is today, but the result of this change makes masterdata deny the request with a 403/401 response even for public entities. Can you test it first @gabrielMatosBoubee ?

The test is very simple: use any public entity on masterdata, save its acronym, go to your site and run`deno task start` remember to link this apps version. To illustrate the failure go to the vtex/mod.ts file and change the headers to any invalid value just to force the failure. See the result when trying to invoke the create document and post it here if possible, 

If this work, so I think we can merge, otherwise we should think in a way that this not become a breaking change.

One option would be receiving a new boolean prop that specifies which client should be used. Maybe "privateEntity?: boolean"; if true, uses vcs, otherwise use deprecated.

guitavano (approved):

guitavano (approved):

PR #371, opened 2024-02-06
[ VTEX ] FIX Canonical in PLP

gabrielMatosBoubee:
# The issue

image

## The fix:
image
image

## The fix support pagination? 
image


guitavano:
I can`t believe, nice catch!

PR #375, opened 2024-02-06
Return videos from VTEX api

teudm:


vtex/utils/transform.ts
```diff
@@ -376,20 +406,7 @@ export const toProduct = <P extends LegacyProductVTEX | ProductVTEX>(
...
-
-      return {
-        "@type": "ImageObject" as const,
-        alternateName,
-        url,
-        name,
-        encodingFormat,
-      };
-    }) ?? [DEFAULT_IMAGE],
+    image: finalVideos ? [...finalImages, ...finalVideos] : finalImages,
```
lucis, line 409:
```suggestion
    image: finalImages.concat(finalVideos || ),
```

lucis (approved):

vtex/utils/transform.ts
```diff
@@ -302,7 +302,8 @@ export const toProduct = <P extends LegacyProductVTEX | ProductVTEX>(
     releaseDate,
     items,
   } = product;
-  const { name, ean, itemId: skuId, referenceId = [], kitItems } = sku;
+  const { name, ean, itemId: skuId, referenceId = [], kitItems, videos } = sku;
+  const video = nonEmptyArray(videos);
```
lucis, line 306: Why `video`? It seems like you're just filtering the empty ones

Maybe `const nonEmptyVideos = ...`

vtex/utils/transform.ts
```diff
@@ -339,6 +340,35 @@ export const toProduct = <P extends LegacyProductVTEX | ProductVTEX>(
...
+    return {
+      "@type": "ImageObject" as const,
+      alternateName,
+      url,
+      name,
+      encodingFormat,
+    };
+  }) ?? [DEFAULT_IMAGE];
+
+  const finalVideos = video && video.map((video) => {
```
lucis, line 358:
```suggestion
  const finalVideos = video?.map((video) => {
```

vtex/utils/types.ts
```diff
@@ -739,7 +739,7 @@ export interface Item {
   unitMultiplier: number;
   modalType: unknown | null;
   images: Image[];
-  Videos: string[];
+  videos: string[];
```
lucis, line 742: Just to check: are you sure about this type change? it's possibly a breaking change if anyone was using or it ever existed

teudm: I'm not sure if anyone would be using it because the API doesn't return any "Videos" (with a capital V), only "videos" (with a lowercase v). Consequently, it wasn't even possible to do anything with this object.

vtex/utils/transform.ts
teudm: You are right

guitavano:
There is something that I am thinking here. If someone is using like:

``images.map((image) => <Image url="url" />)``

Whats is gonna happen with their websites after that change? I think we should receive something at admin called:

``returnVideos: boolean``

What do you think?

lucis:
> There is something that I am thinking here. If someone is using like:
> 
> `images.map((image) => <Image url="url" />)`
> 
> Whats is gonna happen with their websites after that change? I think we should receive something at admin called:
> 
> `returnVideos: boolean`
> 
> What do you think?

makes sense

Can't we have a separate `videos` property also?

teudm:
> > There is something that I am thinking here. If someone is using like:
> > `images.map((image) => <Image url="url" />)`
> > Whats is gonna happen with their websites after that change? I think we should receive something at admin called:
> > `returnVideos: boolean`
> > What do you think?
> 
> makes sense
> 
> Can't we have a separate `videos` property also?

Changing the structure of the product type, which is common across other platforms, to include a separate 'videos' property could potentially introduce compatibility issues?

teudm:
> There is something that I am thinking here. If someone is using like:
> 
> `images.map((image) => <Image url="url" />)`
> 
> Whats is gonna happen with their websites after that change? I think we should receive something at admin called:
> 
> `returnVideos: boolean`
> 
> What do you think?

Done!

guitavano:
> > There is something that I am thinking here. If someone is using like:
> > `images.map((image) => <Image url="url" />)`
> > Whats is gonna happen with their websites after that change? I think we should receive something at admin called:
> > `returnVideos: boolean`
> > What do you think?
> 
> makes sense
> 
> Can't we have a separate `videos` property also?

Lucis, we can. And probably it would be easier for devs deal with.

But I am think about schema.org. There is no video prop, will we have problemas with structured data, for example?

guitavano:
Not specific at GoogleShopping, but I found VideoObject at developers google:
https://developers.google.com/search/docs/appearance/structured-data/video?hl=pt-br#video-object

They not show how to use it on products, but I prefer to separate it now.

guitavano:
continued here: https://github.com/deco-cx/apps/pull/379/files

PR #380, opened 2024-02-08
feat: Ignore case when check for selected filter

lui-dias:
I need to modify the url of the department filters, which includes leaving some parts of it in lowercase. Because of this, some filters are not marked as selected

This PR just ignores the case when it does this check


guitavano (approved):

guitavano:
Can you open again? We have to think more about that.

I think you are doing lowercase at your store, but actually the filter comes capitalized. I am not sure about change only the selected verification.

lui-dias:
Yes, for some reason on the current miess site the url is different from the one returned by the api. The categories and brands in the url are always in lowercase on the current site

I thought about copying the algorithm used to see if the filter is selected, but I thought it would make more sense to do this PR, I didn't imagine it would cause any problems

lui-dias:
And how do I reopen this pr? Do I revert this commit or create a new pr?
image

lui-dias:
#394

PR #386, opened 2024-02-08
Uses import map from deco

mcandeia:
https://github.com/deco-sites/admin/pull/1160


files/loaders/app.ts
```diff
@@ -1,50 +1,111 @@
...
+    if (!/\.tsx?$/.test(path)) {
+      continue;
+    }
+
+    const dataUri = contentToDataUri(
+      join(currdir, path),
+      content,
+    );
+
+    console.log(path, join(currdir, path));
```
tlgimenes, line 104:
```suggestion
    console.log(path, join(currdir, path));
```

tlgimenes (approved):

PR #392, opened 2024-02-12
flat categories to some type of pagetypes

guitavano:
# Part 1
At category pages, we look for the current category to show all below in the three.
image

But, at search, collection, brand pages. The products are in different categories and object comes empty:
image

We could flat all categories here, but that would be weird, because we want to know the name of parent category. 

So, I liked @lui-dias purpose here: https://github.com/deco-cx/apps/pull/368. Bring categories separated:

Collection page before:
image

Collection page after:
image

# Part 2
At collection pages, the category link at filter should start with productClusterIds at map. Even at Vtex, it is necessary:
https://www.miess.com.br/velas/138?map=c,productClusterIds does not works.
https://www.miess.com.br/sex-shop/velas/138?map=c,c,productClusterIds works

This PR verify if request has a productClusterId at map and then deal in a different way.

### DIFF
--- a/vtex/loaders/legacy/productListingPage.ts
+++ b/vtex/loaders/legacy/productListingPage.ts
@@ -190,7 +190,7 @@ const loader = async (
 
   const pageTypes = await pageTypesFromPathname(maybeTerm, ctx);
   const pageType = pageTypes.at(-1) || pageTypes[0];
-
+  console.log(pageType)
   const missingParams = typeof maybeMap !== "string" || !maybeTerm;
   const [map, term] = missingParams && fq.length > 0
     ? ["", ""]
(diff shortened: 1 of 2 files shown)


vtex/loaders/legacy/productListingPage.ts
```diff
@@ -190,7 +190,7 @@ const loader = async (
 
   const pageTypes = await pageTypesFromPathname(maybeTerm, ctx);
   const pageType = pageTypes.at(-1) || pageTypes[0];
-
+  console.log(pageType)
```
tlgimenes, line 193: 👀

tlgimenes (approved):

PR #394, opened 2024-02-13
reopen - feat: Ignore case when check for selected filter 

lui-dias:
Reopen of #380

---

I need to modify the url of the department filters, which includes leaving some parts of it in lowercase. Because of this, some filters are not marked as selected

This PR just ignores the case when it does this check


lui-dias:
@guitavano I spoke to @tlgimenes and we thought it would be better to create a prop to ignore the case instead of putting it directly in the code, because this could affect some stores

Do you think you can merge it?

guitavano (approved):

PR #398, opened 2024-02-14
Wap Integration

IncognitaDev:


guitavano:
Very nice!

As this is a full integration, it would be nice if you can put here two things:

- a demo store working (you probably did one to create)
- a loom explaining in general how it was made and things that we should know

IncognitaDev:
Store: https://deco-sites-storefront-wap.deno.dev/ 

obs: Idk why homepage its as not found as its exists on admin, this starts on the lasts rebases

guitavano:
If you can show loaders working, loader props working, pagination, sorting, cart, checkout.

Say why you had to modify types from commerce, props from SEO.

By the way, you can do in portuguese.

IncognitaDev:
Sorry by the too long video 

https://www.loom.com/share/98693c7099d64ead99316fc4df39184f?sid=e57ee3ee-9a3e-4959-b582-abb251dd4000

IncognitaDev:
after fixes demonstration 

https://www.loom.com/share/78ce811e6f274c71908ef3eabd13c2b9?sid=b1ce2894-7cf6-41b9-a150-51e9dc4402a3

wap/utils/transform.ts
```diff
@@ -0,0 +1,435 @@
...
+          }],
+        }
+      )),
+      author: [{
+        "@type": "Author",
+        identifier: pergunta.email,
+        name: pergunta.nome,
+      }],
+    })),
+    ...selected,
```
guitavano, line 355: I do not understand why ...selected is here, it will replace something you did above. Don't? Do you have an example when it is good?

image

wap/utils/transform.ts
```diff
@@ -0,0 +1,435 @@
...
+    key: filtro.nome,
+    label: filtro.nome,
+    quantity: filtro.valores?.length ?? 0,
+    values: filtro.valores?.map((filterItem): FilterToggleValue => {
+      const filterURL = new URL(baseUrl);
+      filterURL.pathname = new URL(filterItem.url).pathname;
+      filterURL.search = new URL(filterItem.url).search;
+
+      return {
+        quantity: 0,
```
guitavano, line 381: API does not send this value? :(

guitavano:
Did you try to do sitemap?

guitavano:
Very good job!!!!!

wap/utils/transform.ts
IncognitaDev: yes, this is the desired behaviour once some products dosent have variations and your infos its on product root, but when they have the info its on her variations

wap/utils/transform.ts
IncognitaDev: no :/

wap/loaders/productListingPage.ts
```diff
@@ -0,0 +1,140 @@
...
+    ["GET /api/v2/front/url/verify"]({
+      url: url.pathname,
+    }).then((response) => response.json()).catch(() => {}) ?? {};
+
+  if (!busca && !nivel) return null;
+
+  const params: Record<string, string | string[]> = {};
+  url.searchParams.delete("busca");
+  url.searchParams.delete("pg");
+  url.searchParams.forEach((v, k) => {
```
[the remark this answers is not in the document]
IncognitaDev: yes, the problem here its: I could'nt find a pattern betewen every filter, so map all they its realy hard, but I can add the basic filters as panel props, the page prop i just dont add because i dont see a real use case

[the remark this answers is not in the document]
guitavano: Nice, page prop is nott necessary I think. The basic filters should be good, you can open a issue.

PR #402, opened 2024-02-15
[FIX] searchPages in vnda

gabrielMatosBoubee:
In this pr we fix how we verify if is a searchPage or no.

### DIFF
--- a/vnda/loaders/productListingPage.ts
+++ b/vnda/loaders/productListingPage.ts
@@ -104,7 +104,7 @@ const searchLoader = async (
   const sort = url.searchParams.get("sort") as Sort;
   const page = Number(url.searchParams.get("page")) || 1;
 
-  const isSearchPage = url.pathname === "/busca" || url.pathname === "/s" || url.pathname === props.searchPagePath;
+  const isSearchPage = props.searchPagePath ? props.searchPagePath ===  url.pathname : url.pathname === "/busca" || url.pathname === "/s";
   const qQueryString = url.searchParams.get("q");
   const term = props.term || props.slug || qQueryString ||
     undefined;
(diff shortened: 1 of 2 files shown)


guitavano (commented):
I prefer this prop coming from VNDA app, what do you think?

vnda/loaders/productListingPage.ts
```diff
@@ -101,7 +104,7 @@ const searchLoader = async (
   const sort = url.searchParams.get("sort") as Sort;
   const page = Number(url.searchParams.get("page")) || 1;
 
-  const isSearchPage = url.pathname === "/busca";
+  const isSearchPage = url.pathname === "/busca" || url.pathname === "/s" || url.pathname === props.searchPagePath;
```
guitavano, line 107: If url.pathname is undefined, ok verify /busca and /s. But if it is filled, we should look only to thath
```suggestion
  const isSearchPage = props.searchPagePath ? props.searchPagePath ===  url.pathname : url.pathname === "/busca" || url.pathname === "/s"```

gabrielMatosBoubee:
> I prefer this prop coming from VNDA app, what do you think?

You have a point, but I prefer to still this prop here because I think is more UI/UX

guitavano:
Should I say what is my search page in each place that I use this loader? 

> > I prefer this prop coming from VNDA app, what do you think?
> 
> You have a point, but I prefer to still this prop here because I think is more UI/UX

Should I say what is my search page in each place that I use this loader?

gabrielMatosBoubee:
> what is my search page in each place that I use this loader?

I think you need to say it only in the searchPage or some variation. In my mind make more sense select it only when you use

guitavano:
@mcandeia what do you think?

guitavano (approved):

PR #403, opened 2024-02-15
Tavano/remove deco flags

guitavano:
Before:
image

After:
image


guitavano:
Thinking about:

```
if (event.name === "deco") {
  globalThis.window.dataLayer.push({
    event: event.name,
    ...event.params
  });
  return;
}
```

PR #397, opened 2024-02-14
[Linx Impulse]: Add Linx impulse app

vitoUwu:
wip

PR #410, opened 2024-02-19
[Linx Impulse]: Add Linx Impulse integration

vitoUwu:


vitoUwu:
https://github.com/deco-cx/apps/pull/397#discussion_r1494678444

Fixed

vitoUwu:
https://github.com/deco-cx/apps/pull/397#discussion_r1494680113 @mcandeia 

During testing, I noticed that I was getting some errors in the terminal, when I realized that some requests with "_frsh" were causing these errors, I'm not sure what could have caused this.

guitavano:
Very nice!

As this is a full integration, it would be nice if you can put here two things:

- a demo store working (you probably did one to create)
- a loom explaining in general how it was made and things that we should know

vitoUwu:
> Very nice!
> 
> As this is a full integration, it would be nice if you can put here two things:
> 
> * a demo store working (you probably did one to create)
> * a loom explaining in general how it was made and things that we should know

For example, I have https://deco-sites-clovis.deno.dev/ , we use product shelves with the Linx Impulse loader, we use autocomplete in the search bar, and we use click and page view event tracking


https://github.com/deco-cx/apps/assets/32278696/a35f95c8-d9a6-4d78-a105-75220473a8a8

linx-impulse/mod.ts
```diff
@@ -0,0 +1,77 @@
...
+   */
+  secretKey?: string;
+  /**
+   * @title Origin URL
+   * @description Set the origin URL to authenticate the request if not using the secret key
+   */
+  origin?: string;
+  /**
+   * @title Image CDN URL
+   * @description e.g.: https://{account}.myvtex.com/
```
guitavano, line 34: Is myvtex url a good example? Will people use that?

linx-impulse/mod.ts
```diff
@@ -0,0 +1,77 @@
...
+    headers: new Headers({ "Accept": "application/json" }),
+  });
+
+  const chaordicApi = createHttpClient<ChaordicAPI>({
+    base: "https://recs.chaordicsystems.com/",
+    headers: new Headers({ "Accept": "application/json" }),
+  });
+
+  const state = {
+    ...props,
```
guitavano, line 67: Let`s choose which props should be in state? I really appreciate use secretKey and appKey only to create the clients.

linx-impulse/loaders/products/productList.ts
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * @title Category Page
+ */
+interface CategoryPageProps extends BaseProps {
+  /**
+   * @hide
+   */
+  page: "category";
+  /**
+   * @description It is recommended to create a global loader so that the loader is not called more than once
```
guitavano, line 69: Nice!

linx-impulse/loaders/products/productList.ts
```diff
@@ -0,0 +1,249 @@
...
+    .replace(/[\u0300-\u036f]/g, "");
+
+/**
+ * @title Page
+ */
+type Props =
+  | HomePageProps
+  | ProductPageProps
+  | CategoryPageProps
+  | SearchPageProps;
```
guitavano, line 110: Bad say that, but we already have problems doing that, can you change to:
```suggestion
type Props = { props: HomePageProps | ProductPageProps | CategoryPageProps | SearchPageProps;
```

guitavano:
Only to confirm, linx-impulse app will always be used with another main ecommerce integration, right? It's why we do not need to create checkout proxy, for example?

linx-impulse/mod.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * @title Linx Impulse
+ */
+export default function Linx(props: State) {
+  const eventsApi = createHttpClient<EventsAPI>({
+    base: "https://api.event.linximpulse.net/",
+    headers: new Headers({
+      "Accept": "application/json",
+      "User-Agent":
+        "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36",
```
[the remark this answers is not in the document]
vitoUwu: Forgot to remove, fixed

linx-impulse/mod.ts
```diff
@@ -0,0 +1,77 @@
...
+/** @title LINX Impulse */
+export interface State {
+  /**
+   * @title Api Key
+   */
+  apiKey: string;
+  /**
+   * @title Secret Key
+   */
+  secretKey?: string;
```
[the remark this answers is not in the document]
vitoUwu: Fixed!

linx-impulse/mod.ts
vitoUwu: It's the same example used in the Linx integration, stores that use linx + vtex have the cdn url like this

linx-impulse/mod.ts
vitoUwu: According to the Linx Impulse documentation, the apiKey and secretKey are sent in the body in POST requests, and in the query params in GET requests, can I define this when creating the client? The apiKey is a little more public, when integrated via script it is sent to the client side

vitoUwu:
> Only to confirm, linx-impulse app will always be used with another main ecommerce integration, right? It's why we do not need to create checkout proxy, for example?

@guitavano Yes, it is mandatory to have a main integration, be it Linx Commerce, VTEX, etc...

linx-impulse/mod.ts
guitavano: It is possible to add on headers, I checked and should work:

image

guitavano (approved):

PR #413, opened 2024-02-20
[WebSite] - RenderHooks

gabrielMatosBoubee:
Here we can run loaders before middlewares and global sections.

This is what we need to do to use this feature. This is a new loader with renderHooks.

image


mcandeia:
Closed due to another solution on table: use an inline script

PR #436, opened 2024-02-29
Ft/build with precompile

hugo-ccabral:
This PR aims to change the compiler options for jsx to "precompile"
image


website/handlers/router.ts
```diff
@@ -111,34 +111,33 @@ export const router = (
 };
```
hugo-ccabral, line 111: Não mexi no arquivo, foi apenas identação do check

mcandeia (changes requested):

platforms/kubernetes/common/cmds/build.ts
```diff
@@ -215,25 +215,83 @@ async function build() {
...
+
+    if (!hasCompilerOptions) {
+      denoJson.compilerOptions = {};
+    }
+
+    if (
+      !denoJson.compilerOptions.jsx ||
+      denoJson.compilerOptions.jsx !== "precompile"
+    ) {
+      denoJson.compilerOptions.jsx = "precompile";
```
tlgimenes, line 255: if you add compilerOptions.jsx you also need to set jsxImportSource

mcandeia (approved):

tlgimenes (approved):

PR #438, opened 2024-03-04
feat: Use sellerName instead sellerId

lui-dias:
I need the sellerName to set up the jsonLD correctly, but `Offer.seller` returns the sellerId instead of the sellerName

I don't know if replacing sellerId with sellerName is the best way, but I haven't found another way that follows schema.org

### DIFF
--- a/vtex/utils/transform.ts
+++ b/vtex/utils/transform.ts
@@ -525,10 +525,10 @@ const toAdditionalPropertiesLegacy = (sku: LegacySkuVTEX): PropertyValue[] => {
   return [...specificationProperties, ...attachmentProperties];
 };
 
-const toOffer = ({ commertialOffer: offer, sellerId }: SellerVTEX): Offer => ({
+const toOffer = ({ commertialOffer: offer, sellerName }: SellerVTEX): Offer => ({
   "@type": "Offer",
   price: offer.spotPrice ?? offer.Price,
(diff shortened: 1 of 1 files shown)


vtex/utils/transform.ts
```diff
@@ -525,10 +525,10 @@ const toAdditionalPropertiesLegacy = (sku: LegacySkuVTEX): PropertyValue[] => {
   return [...specificationProperties, ...attachmentProperties];
 };
 
-const toOffer = ({ commertialOffer: offer, sellerId }: SellerVTEX): Offer => ({
+const toOffer = ({ commertialOffer: offer, sellerName }: SellerVTEX): Offer => ({
   "@type": "Offer",
   price: offer.spotPrice ?? offer.Price,
-  seller: sellerId,
+  seller: sellerName,
```
lui-dias, line 531:
```suggestion
const toOffer = ({ commertialOffer: offer, sellerId, sellerName }: SellerVTEX): Offer => ({
  "@type": "Offer",
  price: offer.spotPrice ?? offer.Price,
  seller: sellerId,
  sellerName,
```
This may be an option but it doesn't follow schema.org

guitavano:
This week I am going deep thinking if we should add "no schema.org props". It's not about your PR, but the decision will impact here.

PR #454, opened 2024-03-12
feat: Adding FileWidget to widgets.ts

yuriassuncx:

This commit causes a new option to appear for widget imports, giving the possibility of uploading files through deco admin.

I came across a use case where a client needed to upload files with extensions (PDF, docx, etc.) and suggested that this be done through the Deco admin. I noticed that this possibility exists if we import the @format file, which is a type of import that sometimes some Deco devs don't know about.

View in Admin:
image

I had to create a component that took advantage of the file upload functionality. Therefore, I thought it was fair and interesting to add this new typing to Deco widgets.

Below is a photo of the component using this functionality:
image

### DIFF
--- a/admin/widgets.ts
+++ b/admin/widgets.ts
@@ -12,3 +12,8 @@ export type HTMLWidget = string;
  * @format video-uri
  */
 export type VideoWidget = string;
+
+/**
+ * @format file
+ */
+export type FileWidget = string;


guitavano:
Very good incentive. 

With images, we have a widget to save in our database, we are currently changing it to work with other types of file.

Adding @viktormarinho here to follow that.

yuriassuncx:
> Very good incentive.
> 
> With images, we have a widget to save in our database, we are currently changing it to work with other types of file.
> 
> Adding @viktormarinho here to follow that.

thx, @guitavano! this would be very good, because currently it generates a base64 link:
image

guitavano:
Closed, deco.cx will do that, because need admin changes.

PR #457, opened 2024-03-14
Tavano/add segment limit

guitavano:

### DIFF
--- a/vtex/mod.ts
+++ b/vtex/mod.ts
@@ -81,9 +81,9 @@ export interface Props {
   platform: "vtex";
 
   /**
-   * @description Rate limit for unique segment requests
+   * @description Maximum quantity of facets that can be requested in a single request. Default is 10
    */
-  segmentRateLimit?: number;
+  facetsLimit?: number;
 }
 
 export const color = 0xf71963;
--- a/vtex/mod.ts
+++ b/vtex/mod.ts
@@ -79,6 +79,11 @@ export interface Props {
    * @description Use VTEX as backend platform
    */
   platform: "vtex";
+
+  /**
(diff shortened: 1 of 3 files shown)


vtex/mod.ts
```diff
@@ -79,6 +79,11 @@ export interface Props {
    * @description Use VTEX as backend platform
    */
   platform: "vtex";
+
+  /**
+   * @description Rate limit for unique segment requests
```
lucis, line 84: Explain better, I think user's don't call those "segments", it's an internal name

vtex/utils/legacy.ts
```diff
@@ -32,13 +32,24 @@ const PAGE_TYPE_TO_MAP_PARAM = {
...
   term: string,
   ctx: AppContext,
 ) => {
-  const segments = segmentsFromTerm(term);
+  let segments = segmentsFromTerm(term);
   const { vcsDeprecated } = ctx;
 
+  const limit = ctx.segmentRateLimit || DEFAULT_SEGMENT_RATE_LIMIT;
+
+  if (segments.length > limit) {
```
lucis, line 46: Shouldn't you splice this above? Maybe in the beginning of the listing page?

This might introduce inconsistency results (e.g: getting product data one way and page type another)

lucis (approved):

vtex/utils/legacy.ts
guitavano: I don't know, I think we can expected inconsistency behaviours after limit. Maybe return null?

Slicing it from pathname in the beggining of loader looks like too complex, working with map parameter for example.

guitavano:
Closed by inactivity, think again if problem appear.

PR #458, opened 2024-03-14
page-builder

gabrielMatosBoubee:

It's for the vtex day.


guitavano:
Closed because we are not using anymore.

PR #463, opened 2024-03-18
[ Routes ] - Fresh and js routes.

gabrielMatosBoubee:

This fix is about too many routes being called in loaders.

I block routes like: 

```js
"/image/app-apple.png?__frsh_c=707043eaa8a5026a7545d9f02028a85ebf7bdead"
"/sw.js"
```

In this context you only can call js files if this file are in the web-site repository. So, if you want to have a js file like:

"/results.js"

You need to put this file in the assets and use the fresh routes.

### DIFF
--- a/website/handlers/router.ts
+++ b/website/handlers/router.ts
@@ -190,6 +190,12 @@ export default function RoutesSelection(
 
     timing?.end();
 
+    const regexFile = /(\.js)|(_frsh)/gm;
+    if (regexFile.test(req.url)) {
+      return new Response(null, {
+        status: 404,
+      });
+    }
     return await server(req, connInfo);
   };
 }


mcandeia (changes requested):
I think we should not block this. At least at the router level, reasons:

1) it is a completely valid usecase to serve js in a handler without put them into the static directory (we had cases where a site is proxying other fresh sites so it will make it stop working)
2) internal fresh things should be handled by fresh, so if the browser is asking for some js valid or fresh route that does not exists we should fix this in the place where this is being added
3) we are not coupled to fresh, we can use the router for other frameworks like htmx so the underlying framework should be opaque

tip:
regex is the slowest thing in the world, always prefer startswith/endswith/includes over regex. Unless it is too complex to apply those so you can switch to regex


Can you describe the problem so we can think in a solution together? Why did those requests reaches too far in the request path? Shouldn’t they be handled by fresh and asset handler?

PR #468, opened 2024-03-19
add sort to product list when input is id or sku

guitavano:
https://github.com/deco-cx/apps/assets/76620866/22fb4c3f-d7b9-46cb-bfb9-568fab87a17b

### DIFF
--- a/vtex/loaders/intelligentSearch/productList.ts
+++ b/vtex/loaders/intelligentSearch/productList.ts
@@ -158,11 +158,8 @@ const fromProps = ({ props }: Props) => {
 };
 
 const preferredSKU = (items: Item[], { props }: Props) => {
-  let fetchedSkus: string[] = [];
-  if (isProductIDList(props)) {
-    fetchedSkus = props.ids ?? [];
-  }
-  return items.find((item) => fetchedSkus.includes(item.itemId)) || items[0];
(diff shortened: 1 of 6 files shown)


vtex/utils/transform.ts
```diff
@@ -969,3 +972,14 @@ export const toReview = (
     };
   });
 };
+
+export const sortProducts = (
```
tlgimenes, line 976: We can do a O(n) sort instead of a O(n^2log(n)) with an insertion sort like algorithm

guitavano: Can you check now?

vtex/loaders/intelligentSearch/productList.ts
```diff
@@ -156,6 +157,14 @@ const fromProps = ({ props }: Props) => {
   throw new Error(`Unknown props: ${JSON.stringify(props)}`);
 };
 
+const preferredSKU = (items: Item[], { props }: Props) => {
+  let fetchedSkus: string[] = [];
+  if (isProductIDList(props)) {
+    fetchedSkus = props.ids ?? [];
+  }
+  return items.find((item) => fetchedSkus.includes(item.itemId)) || items[0];
```
tlgimenes, line 165: the same in here, O(n) instead of O(n^2)

tlgimenes (approved):

PR #479, opened 2024-03-24
add query path if starts with / and use query if facet is ft

guitavano:


guitavano:
I think it is died. I was using on deco-sites/poc-als

PR #482, opened 2024-03-25
fix:  vtex reviews loader get reviews param & rating count

Vavarine:
## Vtex reviews loader
Previously, an incorrect query parameter was being passed to the /reviews-and-ratings/api/reviews route, along with an erroneous value. This led to confusion as the reviews were incorrectly associated with the SKU rather than the product itself.

## `toReview` function
Based on what I know about how Vtex handles reviews and ratings, I've improved the `toReview` function. Now, the number of ratings matches the number of reviews, following Vtex's unified view on both.


vtex/loaders/product/extend.ts
```diff
@@ -16,11 +16,8 @@ export interface Props {
   products: Product[];
 }
 
-const similarsExt = (
-  products: Product[],
-  req: Request,
-  ctx: AppContext,
-) => Promise.all(products.map((p) => withIsSimilarTo(req, ctx, p)));
+const similarsExt = (products: Product[], req: Request, ctx: AppContext) =>
```
lucis, line 19: There were a lot of code formatting changes

Could you double-check that your editor's format is not overriding Deno's default?

https://docs.deno.com/runtime/manual/tools/formatter

guitavano (approved):

guitavano:
Waiting @gabekerchner opinion here. lgtm

gabekerchner:
@guitavano lgtm too!

PR #489, opened 2024-03-27
Ft/low resources on preview 

hugo-ccabral:
Production without state defined for the site:
image

Production with state defined for the site, higher priority:
image

Preview without state defined for the site:
image

Preview with state defined for the site, we stick with the lower bounds:
image


platforms/kubernetes/actions/deployments/create.ts
```diff
@@ -19,6 +19,7 @@ export interface Props {
   runnerImage?: string;
   siteState: SiteState;
   build?: boolean;
+  production?: boolean;
```
hugo-ccabral, line 22: Pensei em usar o `siteState`

PR #491, opened 2024-03-28
feat: choose whether to concatenate URL parameters in the redirect.

jonathanwlopes:
Greetings,

This PR adds the option for the user to choose whether they want to keep the parameters in the redirect or not.
The functionality was changed to add the feature for files loaded via CSV and through the dashboard.

One issue that is occurring is that in the admin panel, the option, although functional, is not being saved. I would need some assistance or correction for this.

image
* In this image, I demonstrate that despite being active, if I make it public, it will work. However, after refreshing the page, it remains functional but the active status is not retained.

** The parameter combination was initially starting as true by default. In this PR, I've changed it to false because "from" leads to "to," and if the user wishes "from" to lead to "to" with parameters added, they can enable it.

See how it works (PT-BR):
https://www.loom.com/share/24b287d5634c42178c9a522fc20e5c37

### DIFF
--- a/website/handlers/redirect.ts
+++ b/website/handlers/redirect.ts
@@ -12,7 +12,7 @@ export interface RedirectConfig {
  * @description Redirect request to another url
  */
 export default function Redirect(
-  { to, type = "temporary", keepQueryParameters }: RedirectConfig,
+  { to, type = "temporary", discardQueryParameters }: RedirectConfig,
 ) {
   /** https://archive.is/kWvxu */
   const statusByRedirectType: Record<
--- a/website/handlers/redirect.ts
+++ b/website/handlers/redirect.ts
@@ -4,7 +4,7 @@ import { isFreshCtx } from "./fresh.ts";
 export interface RedirectConfig {
   to: string;
   type?: "permanent" | "temporary";
-  concatenateParams?: boolean;
+  keepQueryParameters?: boolean;
 }
 
 /**
@@ -12,7 +12,7 @@ export interface RedirectConfig {
  * @description Redirect request to another url
  */
 export default function Redirect(
-  { to, type = "temporary", concatenateParams }: RedirectConfig,
+  { to, type = "temporary", keepQueryParameters }: RedirectConfig,
 ) {
   /** https://archive.is/kWvxu */
   const statusByRedirectType: Record<
@@ -42,7 +42,7 @@ export default function Redirect(
      *
      * (Useful for tracking parameters e.g Google's gclid, utm_source...)
      */
-    const finalLocation = !queryString || !concatenateParams
+    const finalLocation = !queryString || !keepQueryParameters
       ? location
       : location.includes("?")
       ? `${location}&${queryString}`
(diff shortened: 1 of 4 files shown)


website/handlers/redirect.ts
```diff
@@ -4,13 +4,16 @@ import { isFreshCtx } from "./fresh.ts";
 export interface RedirectConfig {
   to: string;
   type?: "permanent" | "temporary";
+  keepQueryParameters?: boolean;
```
matheusgr, line 7:
```suggestion
  discardQueryParameters?: boolean;
```

Current behavior keeps query parameters if they are present.

We should keep current behavior if this new parameter is `undefined`.

website/handlers/redirect.ts
```diff
@@ -39,7 +42,7 @@ export default function Redirect({ to, type = "temporary" }: RedirectConfig) {
      *
      * (Useful for tracking parameters e.g Google's gclid, utm_source...)
      */
-    const finalLocation = !queryString
+    const finalLocation = !queryString || !keepQueryParameters
```
matheusgr, line 45:
```suggestion
    const finalLocation = !queryString || discardQueryParameters
```

website/loaders/redirects.ts
```diff
@@ -19,6 +17,7 @@ async function getAllRedirects(ctx: AppContext): Promise<Route[]> {
         __resolveType: "website/handlers/redirect.ts",
         to: redirect.to,
         type: redirect.type,
+        keepQueryParameters: redirect.keepQueryParameters,
```
matheusgr, line 20:
```suggestion
        discardQueryParameters: redirect.discardQueryParameters,
```

website/loaders/redirectsFromCsv.ts
```diff
@@ -45,45 +56,39 @@ const getRedirectFromFile = async (
     .slice(1)
     .map((row) => {
       // this regex is necessary to handle csv with comma as part of value
-      const parts = row.split(/,(?=(?:(?:[^"]*"){2})*[^"]*$)/);
-
-      const typeRowIndex = parts.findIndex((part: string) =>
-        REDIRECT_TYPE_ENUM.includes(part)
-      );
+      const parts = row.split(/[,;](?=(?:(?:[^"]*"){2})*[^"]*$)/);
```
matheusgr, line 59: avoid new separators. it may break current users.

```suggestion
      const parts = row.split(/,(?=(?:(?:[^"]*"){2})*[^"]*$)/);
```

matheusgr (changes requested):

guitavano:
@jonathanwlopes can you check @matheusgr comments?

jonathanwlopes:
> @jonathanwlopes can you check @matheusgr comments?

@guitavano Yes, I just completed the verification and adaptation.
@matheusgr You're welcome for the suggestions.

jonathanwlopes:
@matheusgr @guitavano
can you validate if it's better now?

matheusgr (approved):

matheusgr:
> @matheusgr @guitavano can you validate if it's better now?

great work @jonathanwlopes!

proposal: maybe we could add some tests in another PR dealing with redirect. It will improve how we evaluate and discuss about this feature.

PR #495, opened 2024-03-29
Make googlefonts loader cacheable.

matheusgr:

Create a cachekey for googlefonts loader based upon fonts (GoogleFonts[]) property.


igorbrasileiro (approved):
LGTM, but the fonts result from google can change, even the props.fonts still the same?

matheusgr:
> LGTM, but the fonts result from google can change, even the props.fonts still the same?

Cache loader should have a max age of 60s, so, it is okay.

PR #509, opened 2024-04-05
fix: default theme selector

monuelo:
Fix theme selector by adding default theme property


viktormarinho (approved):
Lgtm 👍

PR #533, opened 2024-04-18
VNDA: add order sort to productList when input is id

martinadam75:
Just like this [PR](https://github.com/deco-cx/apps/pull/468), this contribution adds to the **VNDA** `productList` the possibility to choose the order of the products that are displayed via id input.
## Loom
product-order


martinadam75:
@guitavano

guitavano (approved):

PR #535, opened 2024-04-18
Avoid frsh routes

matheusgr:

Avoid router to match /frsh routes (and accept 200 responses)

### DIFF
--- a/website/handlers/router.ts
+++ b/website/handlers/router.ts
@@ -49,9 +49,10 @@ export const router = (
   hrefRoutes: Record<string, Resolvable<Handler>> = {},
   resolver: ResolveFunc,
   configs?: ResolveOptions,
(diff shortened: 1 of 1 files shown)


caroluchoa (approved):

website/handlers/router.ts
```diff
@@ -153,7 +154,12 @@ export default function RoutesSelection(
 ): Handler {
   return async (req: Request, connInfo: ConnInfo): Promise<Response> => {
     // TODO: (@tlgimenes) Remove routing from request cycle
```
lucis, line 156: Shouldn't the TODO be moved to the line below?

matheusgr: It is an overall comment about the route function. Maybe it should be one line above, but we don't have a defined comment style on how should we do it.

PR #539, opened 2024-04-22
Fix: Consider the "noIndexing" property when disabling page indexing

tiagosouzac:
This PR adjusts the SEO components to include the "noIndexing" property when disabling page indexing via the Deco CMS. Currently, the behavior is to maintain indexing if the page has data in the JSON LD, even if the user has selected in the CMS that they do not want to index the page. With the fix, the user's choice in the admin becomes the priority, so the page will not be indexed, even if it has data in the JSON LD.

Old behavior with indexing disabled in the Deco admin:

image

Behavior after the fix:

image


igorbrasileiro (changes requested):
Lint all files

tiagosouzac:
> Lint all files

Sorry. It's done!

igorbrasileiro (approved):

PR #544, opened 2024-04-24
PLP - some products without installments

saymondamasio:
Problem: Some products were not coming with installment information

How it was resolved:
Currently the loader only considers the first item of the product to obtain installment data, but if this product is unavailable it will not be paid in installments. 
I solved it with a function that always looks for the first available item to obtain the installment data.

Thx


guitavano:
I liked, are you using on some store to check?

saymondamasio:
> I liked, are you using on some store to check?

No store is using it, I only tested it at Clovis

saymondamasio:
> Isn't this a param for the VTEX API?

🤔 I don't know, what would it be?

tlgimenes:
Isn't this the hideUnavailableItems from the vtex API? I personally don't like changing this order since some stores may use the exact order for some reason or another. However, if the hideUnavailableItems does not work, I guess this is ok

saymondamasio:
> Isn't this the hideUnavailableItems from the vtex API? I personally don't like changing this order since some stores may use the exact order for some reason or another. However, if the hideUnavailableItems does not work, I guess this is ok

I understand, the problem is that this flag only exists in intelligent search, this adjustment is in legacy

matheusgr (approved):

PR #545, opened 2024-04-24
[ New App ] - Konfidency

gabrielMatosBoubee:

This is about a new app. Konfidency

https://www.embelleze.com/shampoo-novex-hidratante-regenerador-recarga-de-queratina-brasileira-07566/p?skuId=2145371093


mcandeia (approved):

konfidency/loaders/productDetailsPage.ts
```diff
@@ -0,0 +1,54 @@
...
+      return {
+        ...productDetailsPage,
+        product: {
+          ...productDetailsPage.product,
+          aggregateRating,
+          review,
+        },
+      };
+    } catch (error) {
+      console.log(error);
```
mcandeia, line 50: How about use our logger instead?

use 

import { logger } from "deco/mod.ts"; instead

mcandeia: Remember, you need to log a STRING, so you should extract the error message.

konfidency/loaders/productDetailsPage.ts
```diff
@@ -0,0 +1,54 @@
...
+      return { ...productDetailsPage };
```
mcandeia, line 51: why did you need to shallow copy here?

konfidency/loaders/productDetailsPage.ts
```diff
@@ -0,0 +1,54 @@
...
+  ctx: AppContext,
+): ExtensionOf<ProductDetailsPage | null> {
+  const { api, customer } = ctx;
+  return async (productDetailsPage: ProductDetailsPage | null) => {
+    if (!productDetailsPage) {
+      return null;
+    }
+
+    if (!productDetailsPage.product.inProductGroupWithID) {
+      return { ...productDetailsPage };
```
mcandeia, line 29: why did you need to shallow copy here?

konfidency/utils/client.ts
```diff
@@ -0,0 +1,12 @@
+import type { PDPReview } from "./types.ts"
+
```
mcandeia, line 2:
```suggestion
```

mcandeia (approved):
Fix, then shipit

konfidency/loaders/productDetailsPage.ts
gabrielMatosBoubee: To return the productDetails without review. I think is better render the PDP component with or without review.

PR #549, opened 2024-04-25
feat(Product) add complement Name

ApMatheus:
A complement to the product name, information that is defined directly in the Vtex cms when registering the productProduct name complement, information that is defined directly in the Vtex cms when registering the product, with it the business user can insert text that is relevant to the description of the product, such as a collection, highlighting a unique product , among other information that may be relevant


guitavano:
We avoid to use non-schema.org props, can you use this information inside additionalProps?

https://schema.org/

matheusgr (approved):

matheusgr:
We used alternateName on product as the choose property to hold VTEX complement name.

[ApMatheus](https://github.com/ApMatheus)

PR #552, opened 2024-04-26
PLP - Removendo vídeos do JSONLD

saymondamasio:
I have a problem when validating the structured data of a PLP, as you can see below in VTEX they do not bring the video information so there is no error, it is returned in the deco causing the error.

That's why I created a prop in SeoPLP and SeoPLPV2 that removes the videos from JsonLD.

VTEX: https://search.google.com/test/rich-results/result?id=HXln6oT-OVcTSTaf2BdEkA&hl=pt-br
image

Deco:
https://search.google.com/test/rich-results/result/r%2Fvideos?id=B1BdnsQJjDbD44uJomkCLQ&hl=pt-br
image

Thx


guitavano:
Should we really remove video or add thumbnailUrl and updatedDate?

commerce/sections/Seo/SeoPLP.tsx
```diff
@@ -27,6 +36,12 @@ function Section({ jsonLD, ...props }: Props) {
     !jsonLD ||
     !jsonLD.products.length;
 
+  if (props.configJsonLD?.removeVideos) {
+    jsonLD?.products.forEach((product) => {
+      product.video = [];
```
guitavano, line 41: Can you test using undefined?

.vscode/settings.json
```diff
@@ -1,5 +1,7 @@
 {
   "deno.enable": true,
   "deno.lint": true,
-  "deno.unstable": true
+  "deno.unstable": true,
+  "editor.formatOnSave": true,
+  "editor.defaultFormatter": "denoland.vscode-deno"
```
guitavano, line 6: remove that please

commerce/sections/Seo/SeoPLP.tsx
saymondamasio: I changed

.vscode/settings.json
saymondamasio: removed

saymondamasio:
> Should we really remove video or add thumbnailUrl and updatedDate?

unfortunately, vtex does not provide this information

saymondamasio:
I had to remove the SeoPLP config props because the typings were not reflected in the admin, now this feature will only be applied to SeoPLPV2

guitavano (approved):

PR #560, opened 2024-05-01
create loader variants

guitavano:


guitavano:
You have to do things to deal with on admin.

PR #562, opened 2024-05-02
Use single EFS AP for all k8s PVs

pedroteosousa:

We have a hard limit for APs (1000), this change would allow us to have a single AP for all PVs.

### DIFF
--- a/platforms/kubernetes/actions/sites/create.ts
+++ b/platforms/kubernetes/actions/sites/create.ts
@@ -9,6 +9,8 @@ import { k8s } from "../../deps.ts";
 import { AppContext, PREVIEW_SERVICE_SCALING } from "../../mod.ts";
 import { DECO_SITES_PVC } from "../build.ts";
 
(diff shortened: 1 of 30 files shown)


platforms/kubernetes/actions/sites/create.ts
```diff
@@ -117,12 +119,26 @@ export default async function newSite(
   const [secretEnvVar] = await Promise.all([
     secretEnvVarPromise,
     ...setupPromises,
+    corev1Api.createPersistentVolume({
+      metadata: { name: `pv-${siteNs}` },
+      spec: {
+        accessModes: ["ReadWriteMany"],
+        storageClassName: EFS_SC,
+        persistentVolumeReclaimPolicy: "Retain",
```
pedroteosousa, line 127: Not sure if this should be `Retain` or not, I was afraid that if the PV was created and couldn't bind to a PVC, that it would delete the AP. Not sure if that is how that works.

mcandeia (approved):

PR #570, opened 2024-05-06
Magento feature/shipping simulation

JonasJesus42:


magento/utils/cart.ts
```diff
@@ -8,7 +8,7 @@ const ONE_WEEK_MS = 7 * 24 * 3600 * 1_000;
 
 export const getCartCookie = (headers: Headers): string | undefined => {
   const cookies = getCookies(headers);
-  return cookies[CART_COOKIE];
+  return decodeURIComponent(cookies[CART_COOKIE] || "").replace(/"/g, "");
```
devartes, line 11: This causes the get cookies process to become undefined
Error generated if it remains in the format as in pr:
HttpError 400: Missing cart cookie

suggestion:
```
export const getCartCookie = (headers: Headers): string => {
  const cookies = getCookies(headers);

  return `${cookies[CART_COOKIE]}`;
};
```
the cookie needs to follow the format as it is in the store, it needs an encodeURIComponent and then a decodeURIComponent

suggestion:

```
export const setCartCookie = (headers: Headers, cartId: string) => {
  const encodedCartId = encodeURIComponent(`"${decodeURIComponent(cartId)}"`);
  const cookie = `${CART_COOKIE}=${encodedCartId}; Path=/; Expires=${
    new Date(Date.now() + ONE_WEEK_MS).toUTCString()
  }; SameSite=Lax`;
  headers.append("Set-Cookie", cookie);
};
```

JonasJesus42: @devartes This code you showed has a problem, it is sending an "enconded" id to the api which generates an error because it does not accept it, and in the flow if when checking an id it gives an error it falls to catch and creates a new cart, then I'm constantly creating a new cart and saving it to the cookie

and I just realized that there is something strange in this code
` const encodedCartId = encodeURIComponent(`"${decodeURIComponent(cartId)}"`);`
 Why do you decode and hide card ID at the same time? you are activating "" so I have to remove it when picking it up

PR #590, opened 2024-05-16
fix: LegacyItem videos attribute name on PDP loader

pedrobernardina:
On PDP, VTEX Legacy API returns the attribute "Videos", capitalized, whilst Intelligent Search returns "videos". The first case was not being treated, hence products returned by the legacy API would never have videos.

This can be seen in the following product:

https://simplesreserva.vtexcommercestable.com.br/api/catalog_system/pub/products/search/camiseta-manga-curta-algodao-peruano-vermelho-escuro-0077065055/p

image


carcara-gustavo (approved):

viktormarinho (approved):
Lgtm

PR #596, opened 2024-05-17
Cache build routes based on the active/inactive flags

mcandeia:


website/handlers/router.ts
```diff
@@ -166,29 +176,37 @@ export default function RoutesSelection(
...
+        routes: builtRoutes.map((route) => ({
+          pathTemplate: route[0],
+          handler: { value: route[1].func },
+        })),
+        hrefRoutes,
+      };
+    };
+    const routerId = `${RouterId.fromFlags(ctx.flags)}/${ctx.revision ?? ""}`;
+    routerCache[routerId] ??= prepareRoutes();
+    const { routes, hrefRoutes } = routerCache[routerId];
```
matheusgr, line 207: Maybe routerCache should be an weakLFU, since the number of flagIds may have an exponential relation with the number of flags.

Just want to avoid an unbalanced tradeoff between performance and memory.

matheusgr (approved):

PR #603, opened 2024-05-20
feat(records): deco records

igorbrasileiro:


turso/mod.ts
```diff
@@ -0,0 +1,62 @@
...
+export interface Props extends StorageConfig {
+}
+
+/**
+ * @title Turso
+ * @description App for using turso https://turso.tech/
+ * @category Tool
+ * @logo todo
+ */
+export default function Turso(
```
lucis, line 41: Why `Turso` and not `Sqlite`?

lucis (approved):

turso/mod.ts
lucis: And why the two apps and not only one?

records/scripts/pullProd.ts
```diff
@@ -0,0 +1,54 @@
...
+
+async function run() {
+  if (!token || !sitename) {
+    console.error(
+      "DATABASE_AUTH_TOKEN and DECO_SITE_NAME environment variables must be set in order to pull the production database.",
+    );
+    throw error;
+  }
+
+  const productionDumpUrl = `https://${sitename}-decocx.turso.io/dump`;
```
lucis, line 14: [Shouldn't you get this from the app's config?https://www.getlago.com/](https://www.getlago.com/)

records/scripts/pullProd.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  if (!response.ok) {
+    console.error("Could not connect to production database. Make sure your environment variables DATABASE_AUTH_TOKEN and DECO_SITE_NAME are up to date.");
+    console.log("Tip: You can copy a fresh DATABASE_AUTH_TOKEN variable from the Deco.cx admin.");
+    throw error;
+  }
+
+  const dumpQuery = await response.text();
+
+  const command = new Deno.Command("sqlite3", {
```
lucis, line 37: Is this a global exec?

lucis (approved):

records/scripts/pullProd.ts
igorbrasileiro: This runs outside the deco flow. Do you to suggest that should be inside the deco flow?

records/scripts/pullProd.ts
igorbrasileiro: @viktormarinho can you explain this?

PR #617, opened 2024-05-24
Ft/disco bot

hugo-ccabral:


guitavano:
@hugo-ccabral How are we here?

matheusgr:
There is a ongoing effort to make this possible using another approach by @vitoUwu

PR #620, opened 2024-05-25
feat: anthropic app

monuelo:
New Anthropic app
- `actions/stream.ts` - streaming messages
- `actions/code.ts` - code gen (no streaming)


anthropic/actions/code.ts
```diff
@@ -0,0 +1,26 @@
...
+  _req: Request,
+  ctx: AppContext,
+) {
+  if (!messages) {
+    return shortcircuit(new Response("No messages provided", { status: 400 }));
+  }
+
+  const msg = await ctx.anthropic.messages.create({
+    system,
+    model: "claude-3-opus-20240229",
```
lucis, line 20: Should it be passed as a prop? And have all the options on the type?

anthropic/actions/stream.ts
```diff
@@ -0,0 +1,150 @@
...
+    "anthropic-version": "2023-06-01",
+    "anthropic-beta": "tools-2024-05-16",
+    "content-type": "application/json",
+    "x-api-key": ctx.token ?? "",
+  };
+
+  const payload: Anthropic.Beta.Tools.MessageCreateParamsStreaming = {
+    system,
+    messages,
+    model: "claude-3-sonnet-20240229",
```
lucis, line 128: This will be old soon

lucis (approved):

anthropic/actions/code.ts
monuelo: Agree!

PR #634, opened 2024-05-31
feat: add option to aggregate similar products to verified reviews

pedrobernardina:
### Summary

This update introduces new options and fixes missing or incorrect types to help developers replicate the native Verified Reviews experience (VTEX IO App or Script Tag)

### Fixes
- Corrected types that did not align with the API responses.

### Additions
- **`dateCreated` attribute to `AggregateRating`:**  
  - Verified Reviews now display the order date instead of the review's published date.  
- **`stats` attribute to Verified Reviews `AggregateRating`:**  
  - Added to calculate how many customers recommend the product.  
- **Option to aggregate reviews for similar products:**  
  - Final ratings are calculated as a weighted average of all similar products' ratings, based on their respective review counts.  
  - This feature can be tested by enabling the new "aggregateSimilarProducts" option in the Admin and running the loader.  
  - **Use case:** In some segments, product variations are registered as separate products instead of SKUs, making it valuable to aggregate their reviews for a more accurate representation.

OBS: I've closed #633 but there are review comments there (that have already been addressed).


verified-reviews/utils/transform.ts
```diff
@@ -4,23 +4,34 @@ import {
...
+  let weightedRating = 0;
+  let totalRatings = 0;
+
+  Object.entries(ratings ?? {}).forEach(([_, [rating]]) => {
+    const count = Number(rating.count);
+    const value = Number(parseFloat(rating.rate).toFixed(1));
+
+    totalRatings += count;
+    weightedRating += count * value;
+  });
```
gabrielMatosBoubee, line 23: Use reduce: 

```js
const { weightedRating, totalRatings } = Object.entries(ratings ?? {}).reduce(
  (acc, [_, ratingDetails]) => {
    const count = Number(ratingDetails.count);
    const value = Number(parseFloat(ratingDetails.rate).toFixed(1));

    acc.totalRatings += count;
    acc.weightedRating += count * value;

    return acc;
  },
  { weightedRating: 0, totalRatings: 0 }
);
```

gabrielMatosBoubee (changes requested):

matheusgr:
@pedrobernardina Can you apply the @gabrielMatosBoubee suggestion?

guitavano:
Up @pedrobernardina

matheusgr (approved):

PR #645, opened 2024-06-07
change object from we get specifications

guitavano:

### DIFF
--- a/vtex/utils/transform.ts
+++ b/vtex/utils/transform.ts
@@ -494,16 +494,21 @@ const legacyToProductGroupAdditionalProperties = (product: LegacyProductVTEX) =>
     );
   }) ?? [];
 
-const toProductGroupAdditionalProperties = ({ properties = [] }: ProductVTEX) =>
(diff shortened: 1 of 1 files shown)


vtex/utils/transform.ts
```diff
@@ -494,16 +494,21 @@ const legacyToProductGroupAdditionalProperties = (product: LegacyProductVTEX) =>
...
+) =>
+  specificationGroups.flatMap(({ name: groupName, specifications }) =>
+    specifications.flatMap(({ name, values }) =>
+      values.map(
+        (value) =>
+          ({
+            "@type": "PropertyValue",
+            name,
+            value,
+            propertyID: groupName,
```
igorbrasileiro, line 508: Does this property exists in PropertyValue?

igorbrasileiro (approved):

vtex/utils/transform.ts
guitavano: yes

PR #649, opened 2024-06-10
fix: get page type from search params in intelligent search listing page loader

tiagosouzac:
This PR corrects how the page type is determined when using VTEX's Intelligent Search. Currently, the identification relies solely on the URL path, such as `/department`, even if the full URL includes additional parameters (e.g., `/department?filter.category-1=department&filter.category-2=category`). This approach leads to incorrect classification, always identifying the page as a department.

image

The new approach considers the `filter.category` parameters to construct the complete URL, converting, for example, `/department?filter.category-1=department&filter.category-2=category` into `/department/category`. If these parameters are absent, the URL path is used instead.

image

This is beneficial for functionalities that require the accurate page type. For instance, it is useful when setting a specific keyword for the page, as shown in the provided examples.


vtex/utils/intelligentSearch.ts
```diff
@@ -123,3 +124,36 @@ export const setISCookiesBag = (
...
+const segmentsFromTerm = (term: string) => term.split("/").filter(Boolean);
+
+const segmentsFromSearchParams = (url: string) => {
+  const searchParams = new URLSearchParams(url).entries();
+
+  const categories = Array.from(searchParams)
+    .reduce((acc, [key, value]) => {
+      if (key.includes("filter.category")) {
+        acc.push(value);
+      }
```
guitavano, line 137:
```suggestion
  const categories = Array.from(searchParams).sorted()
    .reduce((acc, [key, value]) => {
      if (key.includes("filter.category")) {
        acc.push(value);
      }
```

PR #660, opened 2024-06-14
add color in product options to product additional property

italofg:
add color in product options to product additional property


viktormarinho (approved):
ok

PR #665, opened 2024-06-17
[deferred/Improvement] Seo  

gabrielMatosBoubee:

Now when pass a bot in the website, the bot will receive the links and seo.


guitavano (commented):
Sections are an empty array when display is false, so even bot can't see sections.

guitavano (commented):
It does not work. Because on inline loader level, there are no sections, ``onBeforeResolveProps`` filter them if it is not partials.

You may have what do you want, if ``onBeforeResolveProps`` function receive req and ctx. Maybe.

PR #668, opened 2024-06-18
feat: retrives user document information endpoint created

yuriassuncx:
## What is this PR for?

This Pull Request introduces a new endpoint designed to capture the document (CPF) field of a user. The purpose of this endpoint is to facilitate communication with a Cashback API, allowing the retrieval of cashback values for a user based on their CPF.

Implementation Details
The created endpoint is:

`https://{account}.vtexcommercestable.com.br/api/dataentities/CL/search?email={email}&_fields=document`

This endpoint filters by the user's email and returns their document field, which contains the user's CPF.

Expected Response:

```
[
  {
    "document": "123.456.789-00"
  }
]
```

# Objective:

- CPF Capture: The endpoint allows obtaining the user's CPF, which is necessary for communication with the Cashback API.
- Cashback API Integration: With the user's CPF, it will be possible to query the available cashback values for the user.
- Email Filtering: The search is performed using the user's email, ensuring that the returned CPF corresponds to the provided email.

Please review the changes and, if everything is correct, proceed with merging this Pull Request.

### DIFF
--- a/vtex/utils/openapi/vcs.openapi.gen.ts
+++ b/vtex/utils/openapi/vcs.openapi.gen.ts
@@ -129,9 +129,12 @@ DocumentId?: string
 }
 }
 /**
- * Retrieves user document information.
+ * Retrieves documents' information, while choosing which fields will be returned and filtering documents by specific fields.
  */
-"GET /api/dataentities/CL/search?email=:email&_fields=document": {
+"GET /api/dataentities/:dataEntityName/search": {
+searchParams: {
+params?: string
+}
 response: {
 [k: string]: any
 }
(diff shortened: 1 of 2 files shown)


vtex/utils/openapi/vcs.openapi.gen.ts
```diff
@@ -128,6 +128,14 @@ Href?: string
 DocumentId?: string
 }
 }
+/**
+ * Retrieves user document information.
+ */
+"GET /api/dataentities/CL/search?email=:email&_fields=document": {
```
guitavano, line 134: My feeling is that "CL" should be dynamic and not static and may be the query params email and _fields too.

yuriassuncx: > My feeling is that "CL" should be dynamic and not static and may be the query params email and _fields too.

I agree and I uploaded the fix to make this possible. Can you review?

PR #697, opened 2024-07-02
Fix vtex spec group

aka-sacci-ccr:
"allSpecificationsGroups" field informs which subgroups the properties present in "allSpecifications" belong to. Example:

Captura de Tela 2024-07-03 às 15 43 50

The "allSpecifications" is what matters to deco. The field "allSpecificationsGroups" just tell us wich subgroups these properties belong to, as already said:

Captura de Tela 2024-07-03 às 15 44 14

Captura de Tela 2024-07-03 às 15 44 58

Captura de Tela 2024-07-03 às 15 45 58

Therefore, this change aims to enable the return of the field subgroup, using the "additionalType"


vtex/utils/transform.ts
```diff
@@ -486,14 +486,29 @@ const toBreadcrumbList = (
...
+  });
 
+  return specifications.flatMap((spec) => {
+    const values = (product as unknown as Record<string, string[]>)[spec];
     return values.map((value) =>
-      toAdditionalPropertySpecification({ name, value })
+      toAdditionalPropertySpecification({
+        name: spec,
+        value,
+        additionalType: groupMappings.find((mapping) => mapping.name === spec)
```
guitavano, line 506: https://github.com/deco-cx/apps/pull/645

Can you check this other PR? We add this info in other prop called propertyID. It would be nice keep the same on both functions.

guitavano (approved):

PR #712, opened 2024-07-11
VTEX(Legacy) - Adding "EstimatedDateArrival"

saymondamasio:
I need the vtex pre-sale date, but the deco loader does not return it.

In this PR I made a change to `toProduct()` so that it returns the `estimatedDateArrival` in the `availableDate` property.

I know that the `availableDate` property does not exist at https://schema.org/Product, if it is not possible to add a new property, what is the solution?


guitavano:
Why dont you use the releaseDate? https://schema.org/releaseDate

saymondamasio:
VTEX already returns the `releaseDate`, to not change the schema, I moved it to an `additionalProperty` called `Estimated Date Arrival`.

This only applies to the Legacy API, as this data is not returned in Intelligent Search.

saymondamasio:
@guitavano

PR #716, opened 2024-07-13
Blog: Expose blogpost unique identifier

eduardobbrito:

A blogpost can be completely edited, so the only unique identifier is the created file hash id. This pr exposes this id so it can be used on the BlogPost section.

### DIFF
--- a/blog/types.ts
+++ b/blog/types.ts
@@ -18,6 +18,7 @@ export interface Category {
 }
 
 export interface BlogPost {
+  name: string,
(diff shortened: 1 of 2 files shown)


matheusgr:
I am asking for @monuelo help.

@eduardobbrito How are you using this unique identifier? We want to understand the use case so we can evaluate this PR.

Somo initial thoughts: change the `name` to `id`, and make it as a hidden prop.

PR #722, opened 2024-07-16
RA Trustvox Integration

alexandre-monteiro-ra:
Adding integration to the RA Trustvox platform.


ra-trustvox/components/TrustvoxCertificate.tsx
```diff
@@ -0,0 +1,20 @@
+import { state } from "../mod.ts";
+
+export default function TrustvoxCertificate() {
+  const { enableStaging = false } = state;
```
mcandeia, line 4: I think you're misusing the state, you should not use state `shared` across all app installations otherwise you can have many different types of issues related to concurrency and/or previewing app using different props

mcandeia: Instead, you need to:

Transform this into a section and add a `inline loader` in order to access `AppContext` and then accessing the state from `ctx` there are multiple examples of how to do it in this repository, you can easily find one of them

ra-trustvox/components/TrustvoxProductDetailsRate.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+          .trustvox-widget-rating:hover span.rating-click-here {
+            text-decoration: underline;
+          }
+
+          .ts-shelf-container .ts-shelf-rate {
+            top: 0 !important;
+          }
+        `}
+      </style>
```
mcandeia, line 34: Question @guitavano should we use tailwind here? even outside a site?

ra-trustvox/components/TrustvoxProductReviews.tsx
```diff
@@ -0,0 +1,66 @@
...
+    >;
+  }
+}
+
+export interface Props {
+  page: ProductDetailsPage | null;
+}
+
+export default function TrustvoxProductReviews({ page }: Props) {
+  const { storeId, enableStaging = false } = state;
```
mcandeia, line 18: you should get from `ctx`

mcandeia (changes requested):

ra-trustvox/mod.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @title RA Trustvox
+ * @description RA trustvox reviews.
+ * @category Review
+ * @logo https://raw.githubusercontent.com/trustvox/deco-apps/enhancement/trustvox-app/ra-trustvox/ra-trustvox.png
+ */
+export default function RATrustvox(
+  props: State,
+): App<Manifest, State> {
+  state = props;
```
mcandeia, line 37: This WILL not work for all cases, only when the app runs only once with the same props.

One example when this would fail is when previewing the app, you will change the props for the production request while previewing the app.

State exists for that reason, you should access `ctx` instead

mcandeia (changes requested):

ra-trustvox/components/TrustvoxCertificate.tsx
alexandre-monteiro-ra: This is an example, the stars component is inserted within the product card component, already existing in the customer's store.

image

To use it, the client must import it into the code:

```
import TrustvoxProductDetailsRate from "apps/ra-trustvox/components/TrustvoxProductDetailsRate.tsx";
```

and render wherever you want:

```
<TrustvoxProductDetailsRate productId={product.productID} />
```

For these cases, would you have any idea how I can do it, adapt it to the way you suggested? @mcandeia

mcandeia: Components should be self-contained and should not depend on external data without receiving from props or from context. For instance, you don't have access to ctx inside an island because it is a server-side only data. So you need to drilldown props until it reaches your components, so you either need sections or you need your component to receive the state you need

mcandeia (changes requested):

ra-trustvox/components/TrustvoxCertificate.tsx
alexandre-monteiro-ra: So how can I solve my problem? Considering these cases, will the user import the component directly into the code? I still don't understand how to do this

ra-trustvox/components/TrustvoxProductDetailsRate.tsx
mateusmartinsra: Hey @guitavano, can you help clarify this?

matheusgr: It is ok to not have tailwind here.

matheusgr:
@alexandre-monteiro-ra please update so we can merge.

mcandeia (approved):

PR #731, opened 2024-07-18
feat: add vtex brands extension

vitoUwu:
This PR adds the possibility of extending products to add brand images to intelligent search loaders, currently these loaders do not return this information


vtex/loaders/product/extend.ts
```diff
@@ -1,21 +1,73 @@
...
+const brandsExt = async (
+  products: Product[],
+  req: Request,
+  ctx: AppContext,
+) => {
+  const url = new URL(req.url);
+  const brandKey = new URL("/brands", url.origin);
+
+  const cache = await caches.open(CACHE_NAME);
+  const brandsResponse = await cache.match(brandKey);
```
matheusgr, line 36: avoid `caches` usage

loaders should be cacheable using the cacheKey loader property.

Since this is an extension, you can make a temporary loader that will invoke the brandLoader extension so it will able to cache it (workaround). Or, we need to fix the cacheability of extensions.

matheusgr (changes requested):

PR #733, opened 2024-07-18
WIP | User id clickhouse

guitavano:


matheusgr:
I will work on this PR. I need to create those functions below.

```
export const generateUserId = () => {
  return "";
};

export const generateSessionId = () => {
  return "";
};
```

PR #735, opened 2024-07-19
fix(website/htmx): prevent deferred render visible auxiliary elements

igorbrasileiro:
before website/Deferred:
image

after website/Deferred:
image


igorbrasileiro:
Add this https://github.com/deco-cx/deco/blob/8f8db999b8e73a1c901a4247baced1069f164b0f/runtime/htmx/Bindings.tsx#L13 to this https://github.com/deco-cx/apps/blob/main/htmx/sections/Deferred.tsx#L69

PR #739, opened 2024-07-22
feat: emailjs app

yuriassuncx:
This PR aims to add the EmailJS app (https://www.emailjs.com/) to Deco.cx.

EmailJS allows developers to send emails directly from JavaScript without server code, supporting both client-side and server-side usage. It easily integrates with popular email services such as Gmail and Outlook, offering features to enhance email functionality. In addition to having a free plan of 200 monthly email sends, which is great for testing applications.

https://www.loom.com/share/edd1e561791f40c3ae29170e1d6a3f11

https://github.com/deco-sites/saveyourmind/blob/emailjs_integration/components/ui/RegistrationButton.tsx#L5


emailjs/actions/send.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface Props {
+  /**
+   * @description Template ID of the email
+   */
+  template_id: string;
+  /**
+   * @description Template parameters of the template
+   */
+  template_params?: { [key: string]: string };
```
matheusgr, line 12: Did you test it using our admin?

It seems something that shouldn't work.. O_O

Usually we would do something like: Array<{ name: string, prop: value}>.

If the keys are previously known, it would be better to provide them here.

matheusgr (changes requested):

emailjs/actions/send.ts
yuriassuncx: > Did you test it using our admin?
> 
> It seems something that shouldn't work.. O_O
> 
> Usually we would do something like: Array<{ name: string, prop: value}>.
> 
> If the keys are previously known, it would be better to provide them here.

Yes, I actually tested it and found out that this was why template_params was not showing up in the Admin preview. Thanks for the observation, Matheus, I'll upload the fix soon. The keys can vary according to the user's choice (he's the one who defines them in the emailjs template, so maybe the best option is to make it variable here too)

yuriassuncx: @matheusgr

I made a change that fixes the incompatibility between the Deco.cx admin format and the EmailJS API format for template parameters. The admin requires an array of objects with 'name' and 'value' keys, while the API expects an object with dynamic keys. The `templateParamsObject` function converts the array to the required object format using `reduce`, ensuring API compatibility before making the call.

```ts
const templateParamsObject = payload.template_params?.reduce(
  (acc, { name, value }) => {
    acc[name] = value;
    return acc;
  },
  {} as { [key: string]: string }
);
```

matheusgr:
@yuriassuncx please, rebase so we can merge it.

yuriassuncx:
> @yuriassuncx please, rebase it so we can merge it.

done :)

PR #750, opened 2024-07-29
Tavano/sort by pagetype

guitavano:
Bring the possibility to set different sort depending on pageType.


guitavano:
Trying an approach to create a loader directly on the store.

PR #768, opened 2024-08-07
Return of videos at PDP Legacy

saymondamasio:
About an adjustment in the return of videos in the PDP VTEX Legacy.

This problem occurred because the `isLegacySku()` function does not work correctly.


vtex/utils/transform.ts
```diff
@@ -347,7 +347,7 @@ export const toProduct = <P extends LegacyProductVTEX | ProductVTEX>(
     kitItems,
   } = sku;
 
-  const videos = isLegacySku(sku) ? sku.Videos : sku.videos;
+  const videos = (sku as LegacySkuVTEX)?.Videos ?? sku.videos;
```
mcandeia, line 350: shouldn't be better to fix the `isLegacySku` function rather than discard it?

mcandeia (changes requested):

vtex/utils/transform.ts
saymondamasio: fixed it @mcandeia

guitavano (approved):

guitavano (approved):

mcandeia (approved):

PR #832, opened 2024-09-03
Refactoring Linx duplicated types

Samuel-Brito19:
Working to mitigate unnecessary duplicated types on Linx related types.


viktormarinho:
Once you are happy with the progress here you can tag me @Samuel-Brito19 !

Btw: I don't think you need to update the deco version, seems out of scope for this PR

Samuel-Brito19:
Hi @viktormarinho! I did some modifications trying to solve the problem, but I have some doubts about some cases.
For example, this `Response` interface is present in 3 files but have no use in none of them. Should I take it off?

```
export interface Response {
  Errors: LinxError[];
  IsValid: boolean;
  RefreshBasket: boolean;
  ResponseCallBack: ResponseCallBack;
  SuccessMessage: null;
  Url: null;
  Warnings: unknown[];
}
```
There's other cases similar to that one. Check my changes and please give me a feedback!

viktormarinho:
> Hi @viktormarinho! I did some modifications trying to solve the problem, but I have some doubts about some cases. For example, this `Response` interface is present in 3 files but have no use in none of them. Should I take it off?
> 
> ```
> export interface Response {
>   Errors: LinxError[];
>   IsValid: boolean;
>   RefreshBasket: boolean;
>   ResponseCallBack: ResponseCallBack;
>   SuccessMessage: null;
>   Url: null;
>   Warnings: unknown[];
> }
> ```
> 
> There's other cases similar to that one. Check my changes and please give me a feedback!

@Samuel-Brito19 If there are types without i believe you can delete them. Can you share the other cases? For the Response case it's ok.

viktormarinho (approved):
Really nice! 👍

Issue #823, opened 2024-08-30
Proxy Fails with ?wc-ajax* URLs, Needs Middleware Workaround


yuriassuncx:
#### Description
When attempting to use a proxy to redirect calls that include the `?` character in the URL, such as `/?wc-ajax*`, the proxy fails to process the request. This issue occurs due to how the `?` character is handled during the construction of the target URL within the proxy.

#### Problem Analysis
The `?` character is a special delimiter in URLs that separates the path from the query parameters. When the target URL is manually constructed, incorrect or redundant addition of `?` can cause issues, especially if:
- The original URL already contains a `?`.
- The query string (`qs`) is empty, leading to an unnecessary `?` at the end of the URL.
- The path (`path`) already includes query parameters.

This led to the need to implement a temporary solution with a separate middleware to handle requests containing `wc-ajax` in the URL. This middleware ensures that the requests are correctly redirected to `https://secure.rosakochen.com.br` with the appropriate headers.

#### Temporary Solution Implemented
A middleware was created to intercept requests containing `wc-ajax` in the URL and construct a new URL using the correct headers. The middleware makes the call directly to the desired URL, ensuring that the request is processed correctly. Here is the middleware code:

```typescript
import { AppMiddlewareContext } from "deco/mod.ts";

const HOP_BY_HOP = [
  "Keep-Alive",
  "Transfer-Encoding",
  "TE",
  "Connection",
  "Trailer",
  "Upgrade",
  "Proxy-Authorization",
  "Proxy-Authenticate",
];

export const handler = async (req: Request, ctx: AppMiddlewareContext) => {
  if (req.url.includes("wc-ajax")) {
    const newUrl = new URL(req.url);
    const headers = new Headers(req.headers);
    HOP_BY_HOP.forEach((h) => headers.delete(h));
    headers.set("origin", "https://secure.rosakochen.com.br");
    headers.set("host", "secure.rosakochen.com.br");
    headers.set("x-forwarded-host", "secure.rosakochen.com.br");
    headers.set("Referer", "https://secure.rosakochen.com.br/checkout/");

    const response = await fetch(
      "https://secure.rosakochen.com.br" + newUrl.search,
      {
        headers,
        method: req.method,
        body: req.body,
      },
    );

    return response;
  }

  return ctx.next!();
};
```

#### Final Considerations
[...]

PR #837, opened 2024-09-04
Allow users to specify search params on route path

mcandeia:

Allow users to specify search params on route path to be used by URLPattern.

URLPattern already support search params, we are not using it so far, but we have now included. Users needs to specify following URLPattern [documentation](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern).

Example usecase in the linked issue.

## Issue Link

- Issue: [#823 ](https://github.com/deco-cx/apps/issues/823)

### DIFF
--- a/website/handlers/router.ts
+++ b/website/handlers/router.ts
@@ -99,8 +99,16 @@ export const router = (
     }
 
     for (const { pathTemplate: routePath, handler } of routes) {
+      let url;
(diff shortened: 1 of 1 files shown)


guitavano:
Additional: https://github.com/deco-cx/deco/pull/826/files

guitavano (approved):

Issue #847, opened 2024-09-11
Add posthog as a analytics component


matheusgr:
## Issue Type

Please inform the type(s) of issue(s) you are reporting:

- Feature Request

## Description

Posthog is a well known data analytics tool. It would be nice to provide a simple way to integrate it with your application.

My proposal:
- [ ] Add posthog script head
- [ ] Add posthog to listen to deco events
- [ ] Allow easy integration with elements to capture click/actions (using data-attributes)

The last topic (data-attr) seems like something that we should rethink with a more overall approach (i.e. allow any html element to be tagged via engine).

matheusgr:
Not yet. Only the first two tasks are completed.

matheusgr:
Current posthog implementation makes api as a secret on app, and also, non-optional on sections.

It doesn't need to be a secret (since it is a public key), and should be optional on sections.

PR #848, opened 2024-09-11
Add posthog component

matheusgr:

This PR provides the posthog script and also allow it to listen deco events.

## Issue Link

- Issue: https://github.com/deco-cx/apps/issues/847


guitavano:
I think we should do that on a specific app to posthog like we talked.

PR #863, opened 2024-09-16
[WAKE] previous page button 

lucas-social:
continuation of PR #859 

"previous page" button was replicating the same products on the tested page and did not change the parameter as it repeated the current page


wake/loaders/productListingPage.ts
```diff
@@ -214,7 +214,8 @@ const searchLoader = async (
 
   const hasPreviousPage = page > 1;
 
-  const pageOffset = props.pageOffset ?? 1;
+  // if the home page is not 0, there is a chance that the "previous page" button will replicate the value of the current page
+  const pageOffset = 0;
```
matheusgr, line 218:
```suggestion
  const pageOffset = props.pageOffset ?? 0;
```
You can set the `pageOffset` prop to be zero. It is ok to set the default to zero for this loader but you still need to use this property if needed.

Also, I believe that hasPreviousPage should be: `hasPreviousPage = page > pageOffset`.

matheusgr (changes requested):

matheusgr:
@lucas-social thanks for the changes.

`const hasPreviousPage = page > 1;`

lucas-social:
hey Gaudêncio, I’m still doing some tests but without success. i think it would be interesting to have closer contact since this is affecting my project and could impact its delivery. what do you think?

viktormarinho:
Seems to be working with the current code, but a bug was found in the process of writing the .deco json files. when the json value was 0, deco did not seem to write it. Probably a bug either on our admin or ir our daemon, in some check for a falsy value.

Lucas could follow by now by editing the json directly.

viktormarinho (approved):

Issue #877, opened 2024-09-23
Bug in relatedProductsLoader: Use of Promise.all Causes Loader to Fail When Any productList Call Rejects


portolucas:
**Issue Type**
Bug

### Bug Report
**Description**
There is an issue with the relatedProductsLoader function where it uses Promise.all to fetch related products. If any individual call within the Promise.all fails (e.g., due to an API error or network issue), the entire Promise.all rejects. This causes the loader to fail and propagate the error up the call stack, potentially leading to unhandled promise rejections and application instability.

**Steps to Reproduce (for bugs)**
Use the relatedProductsLoader in an application to fetch related products.
Simulate a failure in one of the batched calls to productList (e.g., by causing an API endpoint to return an error or by interrupting the network connection).

Observe that the Promise.all in the loader rejects due to the failure in one of the promises.
Notice that the loader does not handle the error internally, and the error propagates to the calling function.

**Expected Behavior**
The relatedProductsLoader should handle errors internally to ensure that a failure in one of the batched productList calls does not cause the entire loader to fail. Instead, it should:

- Use Promise.allSettled to wait for all promises to complete, regardless of individual failures.
- Collect and return the successful results.
- Handle or log the errors from failed promises appropriately.
- Ensure that the loader's return type remains consistent (Product[] | null), even when some calls fail.

**Actual Behavior**
- The loader uses Promise.all, which rejects if any promise rejects.
- A failure in any productList call causes the entire loader to reject.
- Errors are not handled within the loader, leading to potential unhandled promise rejections.
- The calling code may receive unexpected errors or null values, disrupting the application flow.

**Additional Context**
This issue affects the robustness and resilience of applications using the relatedProductsLoader.

Modifying the loader to use Promise.allSettled and handling the results can resolve the problem.

**Example modification:**

```
const relatedProductsResults = await Promise.allSettled(
  batchedIds.map((ids) =>
    productList({ props: { similars: false, ids } }, req, ctx)
  ),
);

const relatedProducts = relatedProductsResults
  .filter(
    (result): result is PromiseFulfilledResult<Product[]> =>
      result.status === "fulfilled",
  )
  .flatMap((result) => result.value)
  .filter((x): x is Product => Boolean(x));

```
By handling errors internally, the loader becomes more stable and prevents failures from propagating unnecessarily.

IncognitaDev:
Close by the already merged changes

PR #878, opened 2024-09-23
fix: change related products promise type

portolucas:

Brief Description of Changes:

This pull request enhances the relatedProductsLoader function by improving its error handling mechanism. Previously, the loader used Promise.all to fetch related products in batches, which caused the entire loader to fail if any single productList call rejected due to an error (e.g., API failure or network issue). This behavior led to unhandled promise rejections and potential application instability.

To address this issue, the following changes have been made:

Replaced Promise.all with Promise.allSettled: This allows all promises to complete regardless of individual failures. It prevents the loader from rejecting entirely when a single promise fails.

Result Handling: The loader now filters the fulfilled promises to collect successful results and handles rejected promises by logging the errors. This ensures that the loader returns a consistent type (Product[] | null) without propagating errors unnecessarily.

Error Logging: Added console error logs for rejected promises to aid in debugging and monitoring of failed requests.

These enhancements improve the robustness and resilience of the relatedProductsLoader, ensuring that applications using it can handle partial failures gracefully without experiencing complete breakdowns in functionality.

## Issue Link

- Issue: [#877](https://github.com/deco-cx/apps/issues/877)

### DIFF
--- a/vtex/loaders/legacy/relatedProductsLoader.ts
+++ b/vtex/loaders/legacy/relatedProductsLoader.ts
@@ -119,12 +119,21 @@ async function loader(
 
   relatedProductsResults
     .filter((result) => result.status === "rejected")
-    .forEach((result) =>
-      console.error("Error loading related products:", result.reason)
-    );
+    .forEach((result, index) => {
+      console.error(
+        `Error loading related products for batch ${index}:`,
+        (result as PromiseRejectedResult).reason,
+      );
+    });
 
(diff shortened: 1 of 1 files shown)


matheusgr:
@portolucas Can we review this PR, or there still working to do?

One thing that comes to mind: we should log those errors, so we can monitor/take action if needed.

Also, if all promises fail, maybe we should keep current behavior: this loader should fail.

matheusgr (approved):

PR #884, opened 2024-09-26
fix: Add browser headers to loaders

Gmantiqueira:
For some reason, SAP requests weren't working in staging. We need to check why those headers are needed in staging, because the same requests were working in local environments


matheusgr:
Should we still keep this PR open?

PR #894, opened 2024-10-02
[Shopify] Fix pagination loader PLP

luanargolodev:
This PR aims to fix pagination in the Shopify app's PLP loader.

### What happens?
Technically, when we enter the PLP home page, we would be visiting the first page. The thing is, in the loader, it is inserting the page as 0 and ends up "messing up" the pagination, since the nextPage and previousPage are done from the current pagination.

### Exemplifying
Considering that we enter the /accessories/clothes page and when entering the first page, we do not have the attribute ?page=1, the loader would automatically return that we were on page 0.

When pagination to the next page, it would be set to the url ?page=1 instead of ?page=2. When returning to the pagination, it would be set to the url ?page=0.

image

image


shopify/loaders/ProductListingPage.ts
```diff
@@ -73,7 +73,7 @@ const loader = async (
 
   const count = props.count ?? 12;
   const query = props.query || url.searchParams.get("q") || "";
-  const page = props.page || Number(url.searchParams.get("page")) || 0;
+  const page = props.page || Number(url.searchParams.get("page")) || 1;
```
guitavano, line 76: I think we should bring it optional. Did you see how other integrations are doing that? pageOffSet prop

luanargolodev: thanks, @guitavano.

PR #895, opened 2024-10-02
Opt keep empty paths on http client

matheusgr:

Some httpclients may need empty segments, specially at the end of the URL.

### DIFF
--- a/utils/http.ts
+++ b/utils/http.ts
@@ -85,7 +85,7 @@ export const createHttpClient = <T>(
     headers: defaultHeaders,
     processHeaders = (headers?: Headers) => headers,
     fetcher = fetchSafe,
-    keepEmptyPath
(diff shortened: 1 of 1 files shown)


utils/http.ts
```diff
@@ -76,13 +76,16 @@ export interface HttpClientOptions {
   headers?: Headers;
   processHeaders?: (headers?: Headers) => Headers | undefined;
   fetcher?: typeof fetch;
+  // Keep empty path segments in the URL
+  keepEmptyPath?: boolean;
```
mcandeia, line 80: keepTrailingSlashes?: boolean;

mcandeia (approved):

utils/http.ts
matheusgr: keepEmptySegments translate better this option

PR #912, opened 2024-10-09
feat: create wake checkout metadata actions and add on page load

IncognitaDev:

This PR add wake checkout metadata actions and set this metadata on page load to set utm params on checkout


matheusgr:
@IncognitaDev we missed the review of this task. Is it still valid?

IncognitaDev:
Yeah, this changes is in use by beyoung project with a fork that has this pr merged.

wake/hooks/context.ts
```diff
@@ -135,6 +136,11 @@ if (IS_BROWSER) {
     "visibilitychange",
     () => document.visibilityState === "visible" && enqueue(load),
   );
+
+  const metadata = getUTMMetadata(globalThis.location.search);
+  if (metadata) {
+    await invoke.wake.actions.cart.addMetadata({ metadata });
+  }
```
matheusgr, line 143: This context is no longer used.

matheusgr (changes requested):

PR #917, opened 2024-10-10
[Wake] Feat: add partner token cookie and use in every request that suports It

IncognitaDev:

This PR adds function to set and delete partner token cookie and use in every request that suports this token.
This PR also change the partner prop of wake ProductListingPage to make this prop more meaningful


wake/loaders/productDetailsPage.ts
```diff
@@ -31,8 +32,12 @@ async function loader(
   const { slug, buyTogether, includeSameParent } = props;
   const { storefront } = ctx;
 
+  const partnerAccessToken = getPartnerCookie(req.headers);
+
   const headers = parseHeaders(req.headers);
 
+  console.log({ partnerAccessToken });
```
guitavano, line 39: remove

shopify/utils/storefront/storefront.graphql.gen.ts
```diff
@@ -7704,6 +7704,8 @@ export type FilterFragment = { id: string, label: string, type: FilterType, valu
...
 
+export type CustomerFragment = { id: string, email?: string | null, firstName?: string | null, lastName?: string | null };
+
```
guitavano, line 7708: Should be here? shopify folder?

wake/loaders/productListingPage.ts
```diff
@@ -107,9 +108,10 @@ export interface Props {
   pageHref?: string;
 
   /**
+   * @title Partner Param
    * @description page param to partners page
    */
-  slug?: RequestURLParam;
+  partnerAlias?: RequestURLParam;
```
guitavano, line 114: It brokes the current sites using wake api, right?

guitavano (changes requested):

shopify/utils/storefront/storefront.graphql.gen.ts
IncognitaDev: Maybe the rebase make this, i will remove this

wake/loaders/productListingPage.ts
matheusgr: we may have other clients that may need it.

Can you keep the slug property, tag it as deprecated, and define parterAlias with slug if it is not defined (i.e. `partnerAlias ??= slug`)?

matheusgr (approved):

PR #960, opened 2024-11-18
feat: add date-time type on widgets

yuriassuncx:

This PR introduces a new widget type to the Deco.cx CMS: **DateTime**, which enables the selection and display of both date and time.

- **Type**: `DateTime` (format `string`)

image

With this addition, users can easily include date and time fields in their content structures, providing greater flexibility for managing date-related content.

### DIFF
--- a/admin/widgets.ts
+++ b/admin/widgets.ts
@@ -25,6 +25,11 @@ export type SiteRoute = string;
 export type MapWidget = string;
 
 /**
+ * @format date-time
+ */
+export type DateTime = string;
+
+/**
  * @format textarea
  */
 export type TextArea = string;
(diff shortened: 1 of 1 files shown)


admin/widgets.ts
```diff
@@ -24,6 +24,11 @@ export type SiteRoute = string;
  */
 export type MapWidget = string;
 
+/**
+ * @format date-time
+ */
+export type DateTime = string;
+
```
matheusgr, line 31: Add date as another possible type. Also, add 'widget' to avoid confusion with base types (`Date`).

```suggestion
/**
 * @format date
 */
export type DateWidget = string;

/**
 * @format date-time
 */
export type DateTimeWidget = string;

```

matheusgr (changes requested):

matheusgr (approved):

PR #977, opened 2024-12-18
fix: sku is replaced by product ID

ApMatheus:

In contact with the people at Konfidency, they explained that the review is not sent by the product sku but by the product ID, therefore there is a discrepancy in the documentation 

Below are screenshots of the messages

image
image


IncognitaDev (commented):
LGTM

matheusgr:
We will only change the title to keep the API compatible but make the user aware about this difference.

PR #983, opened 2024-12-26
feat(blog): expose unique identifier

luanargolodev:

This PR aims to add the unique id of the blog post.

As the post can be 100% edited (including the slug), we added this id to have a unique identifier.

Aiming at a scenario in which we save comments using the post_slug in the comments table, if there were any changes to the slug due to an error or anything else, we would lose all data registered in the system due to this change.
{07EE2404-F2E0-4185-B50A-EB3C6DE31511}

### DIFF
--- a/blog/utils/records.ts
+++ b/blog/utils/records.ts
@@ -11,5 +11,11 @@ export async function getRecordsByPath<T>(
   const current = Object.entries(resolvables).flatMap(([key, value]) => {
     return key.startsWith(path) ? value : [];
   });
-  return (current as Record<string, T>[]).map((item) => item[accessor]);
+  return (current as Record<string, T>[]).map((item) => {
+    const id = (item.name as string).split(path)[1]?.replace("/", "");
+    return {
+      ...item[accessor],
+      id,
+    };
+  });
 }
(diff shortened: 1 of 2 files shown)


blog/utils/records.ts
```diff
@@ -11,5 +11,11 @@ export async function getRecordsByPath<T>(
...
     return key.startsWith(path) ? value : [];
   });
-  return (current as Record<string, T>[]).map((item) => item[accessor]);
+  return (current as Record<string, T>[]).map((item) => {
+    const id = (item.name as string).split(path)[1]?.replace("/", "");
+    return {
+      ...item[accessor],
+      id,
+    };
+  });
```
IncognitaDev, line 20: why add it manually? I think this should be in the table, maybe by a workflow that runs it when this changes comes to production

luanargolodev: > why add it manually? I think this should be in the table, maybe by a workflow that runs it when this changes comes to production

@IncognitaDev, by default it already returns, but in a different format.

ex: `collections/blog/categories/05430797e2d1` or `collections/blog/posts/1d76260ef8a8`

the objective is to do the split and return only the unique identifier to use later

luanargolodev: example of how I'm using

{A6300D91-C6DA-4B99-BFC2-5B4070847CD3}

matheusgr (approved):

PR #986, opened 2025-01-07
fix: reviews ordenation

aline-pereira:

The Verified reviews have some custom atributes for some clients. Some clients use date_ASC" others use "date_asc. The PR's suggestion would be to leave a toggle in the admin, so when the client is one that uses the lowercase sorting options, simply activate the toggle.

## Issue Link

- Issue: [#ISSUE_NUMBER](link_to_issue)

## Loom Video

> https://www.loom.com/share/cdd5dadf33304f70a3d6fe3e66f4c736?sid=b2bc9490-ade6-4614-b10b-c43bec0b483c

## Demonstration Link

> https://sites-oficina-reserva--feat-reviews.decocdn.com/camiseta-pima-premium-gola-c-preta-0079110040/p?skuId=8965

### DIFF
--- a/verified-reviews/utils/client.ts
+++ b/verified-reviews/utils/client.ts
@@ -20,15 +20,6 @@ export interface PaginationOptions {
     | "helpfulrating_DESC";
 }
 
-// creating an object to keep backward compatibility
-const orderMap = {
-  date_desc: "date_desc",
-  date_ASC: "date_asc",
-  rate_DESC: "rate_desc",
-  rate_ASC: "rate_asc",
-  helpfulrating_DESC: "most_helpful",
-} as const;
-
 const MessageError = {
   ratings:
     "🔴⭐ Error on call ratings of Verified Review - probably unidentified product",
@@ -96,14 +87,12 @@ export const createClient = (params: ConfigVerifiedReviews | undefined) => {
   };
   /** @description https://documenter.getpostman.com/view/2336519/SVzw6MK5#daf51360-c79e-451a-b627-33bdd0ef66b8 */
   const reviews = (
-    { productId, count = 5, offset = 0, order: _order = "date_desc" }:
+    { productId, count, offset = 0, order = "date_desc" }:
       & PaginationOptions
       & {
         productId: string | string[];
       },
   ) => {
-    const order = orderMap[_order];
-
     const payload = {
       query: "reviews",
       product: Array.isArray(productId) ? productId : [productId],
@@ -122,7 +111,7 @@ export const createClient = (params: ConfigVerifiedReviews | undefined) => {
 
   const fullReview = async ({
     productId,
-    count = 5,
+    count,
     offset = 0,
     order,
   }: PaginationOptions & {
--- a/verified-reviews/utils/client.ts
+++ b/verified-reviews/utils/client.ts
@@ -18,7 +18,17 @@ export interface PaginationOptions {
     | "rate_DESC"
     | "rate_ASC"
     | "helpfulrating_DESC";
+  useOrdenationCompatibility?: boolean;
 }
+
+// creating an object to keep backward compatibility
+const orderMap = {
+  date_desc: "date_desc",
+  date_ASC: "date_asc",
+  rate_DESC: "rate_desc",
+  rate_ASC: "rate_asc",
+  helpfulrating_DESC: "helpfulrating_desc",
+};
 
 const MessageError = {
   ratings:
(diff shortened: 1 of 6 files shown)


verified-reviews/utils/client.ts
```diff
@@ -96,13 +97,13 @@ export const createClient = (params: ConfigVerifiedReviews | undefined) => {
...
   const reviews = (
-    { productId, count = 5, offset = 0, order: _order = "date_desc" }:
+    { productId, count = 5, offset = 0, order: _order = "date_desc", useOrdenationCompatibility = false }:
       & PaginationOptions
       & {
         productId: string | string[];
       },
   ) => {
-    const order = orderMap[_order];
+    const order = useOrdenationCompatibility ? orderMap[_order] : _order;
```
matheusgr, line 106: What is the correct value of sort order?

Is that...   "date_ASC" or "date_asc"? "rate_DESC" or "rate_desc"?

We must use only one value here. The orderMap exists to provide compatibility with previously incorrect values such as date_ASC.

I don't think that we should the user decide which value to send. We should allow only the correct one.

Maybe, we should do... `orderMap[_order] ?? _order` that way, if an user forces the correct ordenation, it will keep working.

matheusgr (changes requested):

matheusgr:
Please provide a description of the problem and what this PR tries to solve.

verified-reviews/utils/client.ts
gabrielMatosBoubee: I think it's because verified reviews have some custom atributes for some clients. So, in that case, we can pass that custom atribute without change the apps and add a "problem"

gabrielMatosBoubee (changes requested):
Show us the problem with some video at loom please?

verified-reviews/utils/client.ts
aline-pereira: Comment updated and loom video added

verified-reviews/utils/client.ts
```diff
@@ -18,6 +18,7 @@ export interface PaginationOptions {
     | "rate_DESC"
     | "rate_ASC"
     | "helpfulrating_DESC";
+  useOrdenationCompatibility?: boolean;
```
matheusgr, line 21: Rename this parameter to: `customizeOrder`.

matheusgr, line 21: Set order as:

```
 order?:
    | "date_desc"
    | "date_ASC"
    | "rate_DESC"
    | "rate_ASC"
    | "helpfulrating_DESC"
    | string;
```

Add a comment explained that 
`"date_desc, date_ASC, rate_DESC, rate_ASC, helpfulrating_DESC"` are legacy compatible parameters that will be mapped to "date_desc, date_asc, rate_desc, rate_asc, helpfulrating_desc"`, but that we overall recommend the usage of a customized parameter.

verified-reviews/utils/client.ts
```diff
@@ -26,8 +27,8 @@ const orderMap = {
   date_ASC: "date_asc",
   rate_DESC: "rate_desc",
   rate_ASC: "rate_asc",
-  helpfulrating_DESC: "most_helpful",
-} as const;
+  helpfulrating_DESC: "helpfulrating_desc",
```
matheusgr, line 30:
```suggestion
  helpfulrating_DESC: "most_helpful" as const,
```

keep it as was before. It may be working for some current client. If needed, the client will customize.

verified-reviews/utils/client.ts
matheusgr, line 21: Add a doc about this parameter (customizeOrder).

verified-reviews/utils/client.ts
matheusgr: Ignore may suggestion. It doesn't work as I expected.

matheusgr (changes requested):

verified-reviews/utils/client.ts
```diff
@@ -96,13 +102,13 @@ export const createClient = (params: ConfigVerifiedReviews | undefined) => {
...
   const reviews = (
-    { productId, count = 5, offset = 0, order: _order = "date_desc" }:
+    { productId, count = 5, offset = 0, order: _order = "date_desc", customizeOrder = false }:
       & PaginationOptions
       & {
         productId: string | string[];
       },
   ) => {
-    const order = orderMap[_order];
+    const order = customizeOrder ? orderMap[_order as keyof typeof orderMap] : _order;
```
matheusgr, line 111:
```suggestion
    const order = customizeOrder ? _order : orderMap[_order as keyof typeof orderMap];
```

as default, keep current behavior.

matheusgr (changes requested):

PR #1001, opened 2025-02-05
[Shopify] - Feat: add shop loader and GraphQL query for shop information

yuriassuncx:
### Motivation  
Shopify already provides store information automatically, and fetching it directly from the API is ideal for building custom pages. This approach eliminates the need to store static data, ensuring that the information remains up to date.  

### What was done  
- [x] Created a loader that queries the Shopify API and retrieves store details, including name, description, and store policies.  
- [x] Added support for fetching specific metafields, allowing greater flexibility for customization.  
- [x] Implemented the `GetShopInfo` query to retrieve relevant data and make it available in the application context. 

image

### Impact  
This improvement makes the Shopify integration more dynamic, enabling future implementations to use store data without manual maintenance. Additionally, it enhances customization by exposing configurable metafields.

### DIFF
--- a/shopify/loaders/shop.ts
+++ b/shopify/loaders/shop.ts
@@ -1,0 +1,33 @@
+import { AppContext } from "../mod.ts";
+import { GetShopInfo } from "../utils/storefront/queries.ts";
+import {
+  Shop,
+  ShopMetafieldsArgs,
+} from "../utils/storefront/storefront.graphql.gen.ts";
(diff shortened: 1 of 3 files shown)


IncognitaDev:
Some of this data may should be used by fields of admin, like the Privacy Policy once this is a content of the privacy policy page, keeping all content management in one place.
Anyway we can go ahead if this PR

PR #1047, opened 2025-03-12
Fix(Website): Check if "src" is undefined before images optimization

vitoUwu:
With this PR the Image component will no more break the entire section if src is not defined correctly.

I'm not happy yet on how we are handling it, now it only shows that are some Image somewhere in my codebase that will not be rendered and I have no clue on where it is. I've tried to use `console.trace` and `console.error(new Error(...))` but the stack trace that I get is not enough to locate where this log is coming from.

tracing

### DIFF
--- a/website/components/Image.tsx
+++ b/website/components/Image.tsx
@@ -166,6 +166,11 @@ const Image = forwardRef<HTMLImageElement, Props>((props, ref) => {
     );
   }
 
+  if (!props.src) {
+    console.error("Missing src. This image will NOT be rendered");
+    return null;
+  }
+
   const srcSet = getSrcSet(props.src, props.width, props.height, props.fit);
   const linkProps = srcSet && {
     imagesrcset: srcSet,


website/components/Image.tsx
```diff
@@ -166,6 +166,11 @@ const Image = forwardRef<HTMLImageElement, Props>((props, ref) => {
     );
   }
 
+  if (!props.src) {
+    console.error("Missing src. This image will NOT be rendered");
+    return null;
+  }
```
gabrielMatosBoubee, line 172: I think we should return some div to find the section with the error:

`<div data-identifier-error />`

website/components/Image.tsx
```diff
@@ -166,6 +166,11 @@ const Image = forwardRef<HTMLImageElement, Props>((props, ref) => {
     );
   }
 
+  if (!props.src) {
+    console.error("Missing src. This image will NOT be rendered");
+    return null;
```
gabrielMatosBoubee, line 171:
```suggestion
    return <div data-identifier-error />;
```

PR #1050, opened 2025-03-13
Improve srcSet flexibility in the Image component

pedrobernardina:

The `Image` component serves as a wrapper that optimizes image URLs for each e-commerce platform while leveraging the optimized Deco CDN. However, the current `srcSet` generation is limited because it only considers two factors, preventing the generation of intermediate image sizes.

Since modifying the existing `srcSet` logic could introduce breaking changes across all stores, a more flexible approach would be to allow an optional `srcSet` prop, enabling customization without altering the default behavior.

## **Proposed Solution:**
Introduce an optional `srcSet` prop to the `Image` component, allowing developers to define custom breakpoints while maintaining backward compatibility.

### DIFF
--- a/website/components/Image.tsx
+++ b/website/components/Image.tsx
@@ -166,7 +166,8 @@ const Image = forwardRef<HTMLImageElement, Props>((props, ref) => {
     );
   }
 
-  const srcSet = getSrcSet(props.src, props.width, props.height, props.fit);
+  const srcSet = props.srcSet ?? getSrcSet(props.src, props.width, props.height, props.fit);
+
   const linkProps = srcSet && {
     imagesrcset: srcSet,
     imagesizes: props.sizes,


gabrielMatosBoubee (changes requested):
I think we should have an array of objects that defines the properties of the image in each viewport.

Ex: 

```jsx
input: const viewPortsConfig  = [
  { viewport: 300, width: 675, height: 1000 },
  { viewport: 600, width: 1200, height: 1800 },
  { viewport: 1024, width: 1920, height: 2880 }
];

output: <img src="https://cdn.exemplo.com/imagens/imagem.jpg?w=675&h=1000" 
     srcset="https://cdn.exemplo.com/imagens/imagem.jpg?w=675&h=1000 675w,
             https://cdn.exemplo.com/imagens/imagem.jpg?w=1200&h=1800 1200w,
             https://cdn.exemplo.com/imagens/imagem.jpg?w=1920&h=2880 1920w"
     sizes="(max-width: 300px) 675px,
            (max-width: 600px) 1200px,
            1920px"
     alt="Imagem otimizada para diferentes dispositivos" />
```

pedrobernardina:
@gabrielMatosBoubee hi!

My proposal was to keep it as simple as possible, but I can see your solution working as a helper function to generate these custom SrcSets — perhaps exported by Image or another file.

PR #1124, opened 2025-04-25
feat: Environment loader

IncognitaDev:

This PR provide a loader that can be used to create props that on development mode can be changed to a env value without need to change the prop configuration like a `Secret` but without need to be encrypted

### DIFF
--- a/website/loaders/environment.ts
+++ b/website/loaders/environment.ts
@@ -1,0 +1,49 @@
+/**
+ * @title Environment
+ * @hideOption true
+ */
+export interface Environment {
+  /**
+   * @ignore
+   */
+  get: () => string | null;
+}
+export interface Props {
+  /**
+   * @title Environment Value
+   */
+  value: string;
+  /**
+   * @title Environment Name
+   * @description Used in dev mode as a environment variable (should not contain spaces or special characters)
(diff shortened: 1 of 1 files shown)


website/loaders/environment.ts
```diff
@@ -0,0 +1,49 @@
...
+   */
+  value: string;
+  /**
+   * @title Environment Name
+   * @description Used in dev mode as a environment variable (should not contain spaces or special characters)
+   * @pattern ^[a-zA-Z_][a-zA-Z0-9_]*$
+   */
+  name?: string;
+}
+const cache: Record<string, string | null> = {};
```
matheusgr, line 23: remove cache. doesnt need it.

website/loaders/environment.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  return cache[value] ??= value;
+};
+/**
+ * @title Environment
+ */
+export default async function Environment(props: Props): Promise<Environment> {
+  const environmentValue = await getEnvironment(props);
+
+  console.log("environmentValue", environmentValue);
```
matheusgr, line 43: remove

```suggestion
```

matheusgr (approved):

PR #1655, opened 2026-08-07
fix: preserve client IP and VTEX route priority when proxying

nicacioliveira:
Two defects in the proxy path, both surfaced while wiring an A/B test between a deco storefront and a VTEX FastStore. Neither is specific to that store — they affect any deco site using `website/handlers/proxy.ts` or the VTEX app.

## 1. `x-real-ip` never reaches the origin

`removeCFHeaders` drops every `cf-*` header, including `cf-connecting-ip`, and nothing replaces it.

Measured on a live pod, proxying 100% of traffic to a header echo and toggling only the `apps/` import:

| `apps` | `x-real-ip` at the origin |
|---|---|
| `0.153.0` | absent |
| this branch | `187.61.224.250` |

`x-forwarded-for` was already fine — it is not in `HOP_BY_HOP` and arrives with the client IP first. Hence the second commit: the first version prepended unconditionally and would have produced a duplicate entry.

## 2. A catch-all `/*` outranks every VTEX system path

Route rank is `(highPriority ? 1000 : 0) + rankRoute(path)`, and `routeFromPath` in `vtex/loaders/proxy.ts` registered without `highPriority`:

```
abTesting  /*         highPriority   1000 + 3 = 1003
vtex       /checkout  —                     6
```

So an A/B audience swallows `/checkout`, `/account`, `/login`, `/api/*`, `/_v/*`, `/arquivos/*` and the rest of `PATHS_TO_PROXY` — in both arms, since routes are ranked before the matcher runs.

The concrete failure: `/checkout` proxies to the FastStore, whose checkout route is only

```js
useEffect(() => { window.location.href = config.checkoutUrl }, [])
```

with `checkoutUrl` hardcoded to the store origin. That redirects straight back into the catch-all. Infinite loop.

Marking those paths `highPriority` puts them at `1006`, ahead of any catch-all.

## Notes

- `removeCFHeaders` is left untouched; it is exported and used elsewhere (`linx/utils/headers.ts`).
- Downstream CDNs append their own hop to `x-forwarded-for`, so origins should read the first entry.
- `deno fmt` and `deno lint` pass. `deno check` reports 2 pre-existing errors in `website/utils/crypto.ts`, identical on a clean `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved proxy IP handling for forwarded requests, including IPv4 and IPv6 addresses with port notation.
  - Preserved the connecting client’s IP and set consistent forwarding headers.
  - Ensured built-in system routes take precedence over configured extra paths.


nicacioliveira:
Validated end to end on a live deco site (`deco-sites/nicacio`), proxying 100% of traffic to a header echo, toggling only the `apps/` import between `0.153.0` and this branch. Same site, same client, same target.

**Before — `apps@0.153.0`**
```
X-Real-Ip        (absent)
X-Forwarded-For  187.61.224.250,10.3.1.165,10.3.120.66, …
```

**After — `apps@73966ef`**
```
X-Real-Ip        187.61.224.250
X-Forwarded-For  187.61.224.250,10.3.1.165,10.3.120.66, …
```

The client IP now reaches the origin via `x-real-ip`, and `x-forwarded-for` is unchanged — one entry for the client, no duplicate. That second point is why the branch has two commits: the first version prepended unconditionally, and a live probe showed `x-forwarded-for` already arrives with the client IP first, so it would have produced `187.61.224.250, 187.61.224.250, …`.

Site was restored to `0.153.0` after the run.

nicacioliveira:
Addressed both cubic findings in `6f69aefe`.

**P3 (dedup comparison) — valid, fixed.** The guard compared raw strings, so IPv6 hex casing differing between hops, or an `x-forwarded-for` entry carrying a port, would slip past it and prepend a duplicate. Now both sides are compared in canonical form (lowercased, brackets and port stripped). The forwarded value itself is untouched — normalization is for comparison only.

```
1.2.3.4:56789      -> 1.2.3.4
[2001:DB8::1]:443  -> 2001:db8::1
2001:DB8::1        -> 2001:db8::1
```

**P2 (trust boundary) — real, but not introduced here.** `x-forwarded-for` is not in `HOP_BY_HOP` and was already forwarded untouched before this PR. An origin reachable outside the CDN could always be fed a forged first entry; deriving `x-real-ip` from `cf-connecting-ip` gives an attacker no capability they did not already have by setting `x-forwarded-for` directly.

Validating against Cloudflare IP ranges is also not enforceable at this layer: the handler runs behind the service mesh, so the peer address it would check is an internal hop, not the edge. Authenticating the edge belongs at the ingress — that is CDN-to-origin auth (mTLS, a shared secret header, or an allowlist), not a header check inside a proxy loader.

Documented the boundary in the code comment as suggested, rather than implementing a check that would look like a guarantee without being one.

nicacioliveira:
Field note for anyone using this alongside an A/B test against a **FastStore** target: `/api/*` is contested and the priority bump alone is not enough.

```
                              FastStore target     legacy platform
/api/graphql                  serves it            no
/api/checkout/pub/orderForm   404                  serves it
/api/sessions                 404                  serves it
```

Reproduced live: with `/api/*` correctly winning over the A/B catch-all, the checkout works but every product listing fails — the FastStore’s `/api/graphql` is now routed to the legacy platform.

Not a defect in this PR: sending `/api/*` to the VTEX platform is right for a deco storefront, and the collision only exists when the A/B target is itself a VTEX app answering under `/api`. The site-level fix is one more route, which outranks it naturally:

```
rankRoute("/api/graphql") = 9   -> 1009 with highPriority
rankRoute("/api/*")       = 6   -> 1006
rankRoute("/*")           = 3   -> 1003
```

```jsonc
{
  "pathTemplate": "/api/graphql",
  "highPriority": true,
  "handler": { "value": {
    "__resolveType": "website/handlers/proxy.ts",
    "url": "https://<faststore-host>"
  }}
}
```

Worth calling out in the A/B docs — without it the symptom is a working checkout and empty PLPs, which reads like a data problem rather than a routing one.

